PR #30, opened 2023-05-18
I18N SlotsGroup

boholder:
Finish [TODO](https://github.com/khanshoaib3/minecraft-access/blob/b6702826583a9580dd9b57da88a68fe1891f5990/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java#L54) about I18N SlotsGroup, following with https://github.com/khanshoaib3/minecraft-access-i18n/pull/2

### DIFF
--- a/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
+++ b/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
@@ -17,7 +17,6 @@ import net.minecraft.inventory.CraftingInventory;
 import net.minecraft.inventory.CraftingResultInventory;
 import net.minecraft.inventory.SimpleInventory;
 import net.minecraft.item.ItemGroup;
-import net.minecraft.item.Items;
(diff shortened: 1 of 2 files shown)


khanshoaib3 (approved):

boholder:
Wait wait wait, some slot groups (like amour group, it's groupName is `Armor`, not `armor`) doesn't work! I'll fix it ASAP

Sometime it works, sometime it doesn't work. I don't know why, this method 

https://github.com/khanshoaib3/minecraft-access/blob/8e1e3bd874d33c10dd56708883547cc81b7bd4b8/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java#L48

should be called in every tick and so it does.

https://github.com/khanshoaib3/minecraft-access/blob/8e1e3bd874d33c10dd56708883547cc81b7bd4b8/common/src/main/java/com/github/khanshoaib3/minecraft_access/MainClass.java#L105-L107

boholder:
Found the reason: there are some `SlotsGroup` instances are initialized with non-config-key groupName

Issue #3, opened 2023-01-09
Speak hovered item 


khanshoaib3:
Speak the hovered item like it used to before

boholder:
This function has been implemented in inventory control.

PR #32, opened 2023-05-18
Reset previousQuery after every 5000ms

boholder:
Solve a related [TODO](https://github.com/khanshoaib3/minecraft-access/blob/b6702826583a9580dd9b57da88a68fe1891f5990/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/ReadCrosshair.java#L54).

BTW, there are some `toSpeak` variables in `ReadCrosshair.class` that marked as *unused*, is that normal?

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,8 @@ redstone signal or when a door is open and other similar cases.
 1. Enabled = Enables this feature.
 2. Speak Block Sides = Enables speaking of the side of block as well.
 3. Disable Speaking Consecutive Blocks With Same Name = Disables speaking the block if the previous block was also same.
+4. Repeat Speaking Interval (in milliseconds) = Repeat speaking the block you are facing for the given amount of time, even you haven't moved the camera.
+   Zero for tuning it off, default: 5000.
 
 ## Position Narrator
 
--- a/common/src/main/java/com/github/khanshoaib3/minecraft_access/config/ConfigMap.java
+++ b/common/src/main/java/com/github/khanshoaib3/minecraft_access/config/ConfigMap.java
@@ -137,6 +137,7 @@ public class ConfigMap {
         defaultReadCrosshairConfigMap.setEnabled(true);
         defaultReadCrosshairConfigMap.setSpeakSide(true);
         defaultReadCrosshairConfigMap.setDisableSpeakingConsecutiveBlocks(true);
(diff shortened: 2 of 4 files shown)


khanshoaib3:
The `toSpeak` and `currentQuery` have different values, `currentQuery` is checked to not speak the same block repeatedly but two blocks can have same name, and should be spoken if the focus is changed. That's why currentQuery is used for checking but `toSpeak` is actually the one to be spoken. Maybe have the `speakIfFresh()` also accept `toSpeak` as parameter and use that in `speakWithNarrator()`.

boholder:
This modification is quite... noisy (for me), the narrator keep reporting about the block you're pointing at every five seconds.
So I make the interval as a config.

boholder:
BTW... Has your development environment's narrator ever failed? 
After a restart of debug mode, my narrator (Windows build-in narrator, not the tolk) stopped talking, even with the corresponding settings turned on. 
I'm not sure what happened or how to fix it.

khanshoaib3:
I mainly use linux so I don't know and I haven't faced anything similar to that on linux, what happens when you refresh the screen reader after it happens?

khanshoaib3 (approved):

boholder:
> what happens when you refresh the screen reader after it happens?

I... can't find a way to refresh the Windows build-in narrator, but things won't be solved even I reboot the computer. Thanks for your information, I'll try to find a way to refresh the narrator or just install the tolk...

khanshoaib3:
I don't think we can refresh the default narrator, I thought you were using tolk.
Maybe the narrator gets disabled, try control + b to change the narrator's mode.

boholder:
> I don't think we can refresh the default narrator, I thought you were using tolk.

I found the reason: [Merge pull request #29 from khanshoaib3/auto-installer](https://github.com/khanshoaib3/minecraft-access/commit/b6702826583a9580dd9b57da88a68fe1891f5990#diff-9cefaa1d613cc32c11bda90d860105294ae477e5b5b19d1c3b983ae87d76205b)
It seems that tolk binaries hijack the narrator-invokes...... After revert the commits, Windows narrator works again. (I'll keep them inside stash and would not commit the revert.) 👌

khanshoaib3:
Yeah, the mod auto installs the tolk library as it supports NVDA, JAWS, the windows narrator and others.

khanshoaib3:
This will probably be fixed when #3 is implemented.

PR #52, opened 2023-06-22
Add speak harvest of fishing feature, to be clear it will report what your picked up when you're holding a fishing rod. 

### LINKED ISSUE #48: Speak the harvest of finishing
Feature request from Discord user lexi/scarlet#7459

implementation idea: mixin somewhere, if holding fish rod, speak the item.
like this one: https://github.com/FabricMC/fabric/issues/1130

boholder:
close #48

### DIFF
--- a/common/src/main/java/com/github/khanshoaib3/minecraft_access/mixin/ClientPlayNetworkHandlerMixin.java
+++ b/common/src/main/java/com/github/khanshoaib3/minecraft_access/mixin/ClientPlayNetworkHandlerMixin.java
@@ -1,0 +1,59 @@
+package com.github.khanshoaib3.minecraft_access.mixin;
+
+import com.github.khanshoaib3.minecraft_access.MainClass;
+import net.minecraft.client.MinecraftClient;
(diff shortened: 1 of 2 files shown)


khanshoaib3:
By looking at the code, I think that it will also speak the item name if we pick up items laying on the ground while holding the fishing rod. I;m not sure though.
Does that happen?

boholder:
> By looking at the code, I think that it will also speak the item name if we pick up items laying on the ground while holding the fishing rod. I;m not sure though. Does that happen?

Yep, that's because the item entity is spawned at the position of the fishing bobber and flies to the player, normally the item will be picked up by the player, same as picking up items on the ground.
I can't find a way to detect "if the player is finishing", because at the time of player picks up item, the state of the player is just a player holding a fishing rod.

khanshoaib3:
Thought so, it's fine for now I guess. Something is better than nothing.
I think we should add a config to make it toggleable.

khanshoaib3 (approved):

boholder:
> I think we should add a config to make it toggleable.

Mm-hmm, I think so, I'll make one.

PR #58, opened 2023-06-23
Refactor logic about feature execution delays (intervals of repeating)

boholder:
Solve TODO mentioned in https://github.com/khanshoaib3/minecraft-access/issues/55

PR #59, opened 2023-06-24
Add marked poi scan feature

### LINKED ISSUE #43: Marked POI Scan Button
Press `Enable Key` to indicate a/an block/entity through crosshair (similar to POI locking) and let the mod continually scan surroundings and report the targets location (similar to fluid detection). Stop the scanning by pressing `Alt` + `Enable Key`.

One thing I'm concerned about is that I remember fluid detection causing a brief lag (about seconds) (default range is 10 blocks). But this is basically a simple BFS algorithm and I can't find any room for optimization.
I don't know if MC allows us to create our own thread pools so we can put this logic in a thread pool. And I don't think it's a good idea to manage threads manually either.

boholder:
close #43 

Hey @khanshoaib3 , you may want to directly merge this PR since #58 and #56 has some conflict between them, and this PR has solved it.

And I tried to convert some `ConfigMap`s into a singleton pattern, and you might be interested in the code: https://github.com/khanshoaib3/minecraft-access/commit/991215ae46b11568e2a5fde3a5b4aa628237d2ea

This PR contains quite a much code, so there is no need to be hurry to merge it, let's review slowly. Let me know where you can not figure it out.  I'd also like to hear from @onion108 , if that's convenient for you.

PS: One thing is a bit strange for me: When the marking is on, **sometimes** marked entities are not detected by `POIEntities`. I'm sure it's not a problem with the code, it seems like these pre-conditons are not filled.

https://github.com/boholder/minecraft-access/blob/aea17edc237199fd6a56c420c65ff469bee90332/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/point_of_interest/POIEntities.java#L76-L87


khanshoaib3 (approved):

khanshoaib3:
That may be because of the distance of the entity. Other conditions should not cause this.

boholder:
> That may be because of the distance of the entity. Other conditions should not cause this.

That's the weird thing it is: the animals are just by my side. I've tried to test if it related to every single entity, but the result seems has no clue.

PR #61, opened 2023-06-27
Fix some things so both Fabric and Forge client can at least start without crush.

boholder:


khanshoaib3 (approved):

khanshoaib3:
Cool, I was about to do these changes but this saves some time!

PR #76, opened 2023-07-06
Set the partial speaking disabled by default; Fix mod crashing on loading wrong config.json;

khanshoaib3:


boholder:
Ok, got it about using config maps from MainClass. Is that exception a `NullPointerException`? Maybe I can add some code in `getInstance()` to prevent that? @khanshoaib3 

Have tried, failed, so let's keep in mind about this while changing other config menus......

Issue #60, opened 2023-06-25
1.20 Update


khanshoaib3:
Update the mod to support minecraft 1.20.x

khanshoaib3:
Affected and working features-
- [x] Fluid Detector 
- [x] Read Crosshair - Regular signs and hanging signs 
- [x] Inventory Controls - Deselect/Unfocus Text fields. Thanks to @boholder it works now!
- [x] Time of Day
- [x] POI Blocks Locking  - ~absolute positions for blocks like levers, comparator, buttons, etc. are not taken into account because the `new BlockPos()` now takes `Vec3i` instead of `Vec3d`. So now the locking feature locks on to the center of every block~ Works! Thanks to @boholder 
- [x] Recipe section buttons
- [x] Book screen
- [x] Book edit screen

boholder:
I'm not sure which part of code is corresponding to `Inventory Controls - Deselect/Unfocus Text fields` inside `InventoryControls.class`...... Did you mean these code:
https://github.com/khanshoaib3/minecraft-access/blob/f615c30e3e6d80f200277d0489e8b2dcdd872b68/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/InventoryControls.java#L177-L214

khanshoaib3:
Yup, these are the ones and I should've replied earlier, I don't know how I missed your message.

boholder:
> Yup, these are the ones and I should've replied earlier, I don't know how I missed your message.

Maybe that's because I didn't quote your reply, I don't know.

PR #82, opened 2023-07-13
Fix broken special block locking feature in 1.20.x

boholder:
You won't believe how easy it is to be solved, I know there must be a method for mojang devs themself to converting old format (`Vec3d`) into new format (`Vec3i`), and there does is!

The reason why it's broken is that the `ImmutableSet<Map.Entry<Property<?>, Comparable<?>>> entries = blockState.getEntries().entrySet();` will always be empty, so I realized we've got the wrong position from `client.world`.

I've tested all five special blocks in `AbsolutePositions.class` to make sure all of them works.


boholder:
Link this PR to #60

khanshoaib3:
ooo so the `ImmutableSet<Map.Entry<Property<?>, Comparable<?>>> entries = blockState.getEntries().entrySet();` was checking the block state of the empty/air block that contained the levers/buttons/etc. This is crazy XD

khanshoaib3 (approved):

PR #83, opened 2023-07-13
Close the narrator menu if the F4 key is pressed while the menu is opening

boholder:


boholder:
~~The conditions of which action should be taken, has grown complexity. It can be simplified into a finite-state machine, I'll solve it  as a little fun challenge.~~ The state machine implementation will has more code than current, so I give up on it.

khanshoaib3 (approved):

khanshoaib3:
Wow, I was thinking about adding this!

PR #85, opened 2023-07-14
Fix book edit screen tab switching buttons feature in 1.20.x

boholder:
The `Enter` key will trigger the sign action while screen is in `signing` state, so I removed it.
Add a config text: https://github.com/khanshoaib3/minecraft-access-i18n/pull/13

### DIFF
--- a/common/src/main/java/com/github/khanshoaib3/minecraft_access/mixin/BookEditScreenMixin.java
+++ b/common/src/main/java/com/github/khanshoaib3/minecraft_access/mixin/BookEditScreenMixin.java
@@ -2,13 +2,16 @@ package com.github.khanshoaib3.minecraft_access.mixin;
 
 import com.github.khanshoaib3.minecraft_access.MainClass;
 import com.github.khanshoaib3.minecraft_access.utils.KeyUtils;
+import com.github.khanshoaib3.minecraft_access.utils.MouseUtils;
 import net.minecraft.client.MinecraftClient;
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/github/khanshoaib3/minecraft_access/mixin/BookEditScreenMixin.java
```diff
@@ -37,57 +41,77 @@ public class BookEditScreenMixin {
     @Shadow private ButtonWidget doneButton;
 
     @Shadow private String title;
-    boolean firstTimeInSignMenu = true;
-    String previousContent = "";
+
+    @Unique boolean minecraft_access$firstTimeInSignMenu = true;
```
khanshoaib3, line 45: What does the `@Unique` does? Is it like unique for each instance or object of the class?

khanshoaib3 (approved):

common/src/main/java/com/github/khanshoaib3/minecraft_access/mixin/BookEditScreenMixin.java
boholder: > What does the `@Unique` does? Is it like unique for each instance or object of the class?

That's a mark for notifying Mixin DO NOT override them onto vanilla code.
After I upgraded some plugins in IntelliJ, these linter  (suggestions) poped up in MIxin classes.

PR #97, opened 2023-08-01
Speak monster spawner mob type

boholder:
Request from Trix from Discord

### DIFF
--- a/common/src/main/resources/minecraft_access-common.mixins.json
+++ b/common/src/main/resources/minecraft_access-common.mixins.json
@@ -4,7 +4,8 @@
   "package": "com.github.khanshoaib3.minecraft_access.mixin",
   "compatibilityLevel": "JAVA_17",
   "mixins": [
-    "ChatHudAccessor"
+    "ChatHudAccessor",
+    "MobSpawnerLogicAccessor"
   ],
   "client": [
     "AnimatedResultButtonAccessor",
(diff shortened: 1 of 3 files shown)


common/src/main/resources/minecraft_access-common.mixins.json
```diff
@@ -4,7 +4,8 @@
   "package": "com.github.khanshoaib3.minecraft_access.mixin",
   "compatibilityLevel": "JAVA_17",
   "mixins": [
```
khanshoaib3, line 6: Just a small refactor, these are better to be moved to "client" section

khanshoaib3 (approved):

common/src/main/resources/minecraft_access-common.mixins.json
boholder: I've noticed that, but I can't find a valiable field for extracting the mob name from pure-client-side classes.

boholder: > Just a small refactor, these are better to be moved to "client" section

I've noticed that, but I can't find a valiable field for extracting the mob name from pure-client-side classes.
Oh, you mean we can simply move the two to `client` list without any error?

khanshoaib3: Yup, that should not cause any error and ensure that the mod is client sided.

PR #107, opened 2023-08-04
Optimize screen groups order, in the order of interactive frequency, …

boholder:
…for reducing the number of user keystrokes.

* Inventory groups first, you always start a process with picking up items.
* Then the input and output groups, you may want to put picked items into the input slots and get the result from output slots.
* Then the non-item-related groups you want to interact with (after you put items into input slots, enchant for example).
* Finally, the armour groups, low interactive frequency, will only show in inventory screen.


khanshoaib3 (approved):

khanshoaib3:
Oh yeah, this will definitely reduce the key strokes and make the process a little less tedious.

Issue #123, opened 2023-08-15
Add `Look Straight Up` and `Look Straight Down` keys to `Camera Controls`


boholder:
* `Look Straight Up` = Snaps the camera to the look above head direction.
* `Look Straight Down` = Snaps the camera to the look down at feet direction.

PR #133, opened 2023-08-19
[1.20] Camera look straight up down FINISH

boholder:
Please merge #132  first for less review burden.

* Add new `DoubleClick` keystroke checker util for this feature's requirement.
* Finish #123

PR #135, opened 2023-08-19
[1.19] Camera look straight up down

boholder:
Same content as #131 #132 #133 , but since the `Book Edition` feature (fixed in 1.20 #85) isn't fixed in this 1.19 branch, will not port related code.


khanshoaib3 (approved):

boholder:
Close #123

PR #280, opened 2024-04-22
fix: add missing annotation resolver dependencies that makes test failed

boholder:
## Describe what you have changed in this PR

Sorry for hurrying PR without checking CI result.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -29,7 +29,10 @@ dependencies {
 
     // lombok that makes the life easier
     // https://projectlombok.org/setup/gradle
-    compileOnly group: "org.projectlombok", name: "lombok", version: "1.18.30"
+    compileOnly 'org.projectlombok:lombok:1.18.30'
+    annotationProcessor 'org.projectlombok:lombok:1.18.30'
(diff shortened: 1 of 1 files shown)


khanshoaib3:
@boholder I'm at fault here too. I noticed the checks failing but thought that was due to the mod not being updated to the latest version. I forgot this is unlike stardew and things don't break that way :(

khanshoaib3 (approved):

khanshoaib3:
@boholder BTW what are your thoughts on keeping the main branch (1.20 atm) linear? Like the dev branch on stardew access. Currently merging directly with GH results in an extra merge commit because it doesn't do fast-forward. I added a workflow on sv access repo that merges the branch with fast-forward. We could use that if we want a linear history.

boholder:
> @boholder BTW what are your thoughts on keeping the main branch (1.20 atm) linear? Like the dev branch on stardew access. Currently merging directly with GH results in an extra merge commit because it doesn't do fast-forward. I added a workflow on sv access repo that merges the branch with fast-forward. We could use that if we want a linear history.

I'll check the sv repo, I sort of get what you want to say but let me confirm that.

PR #295, opened 2024-06-23
Fixed detection of modded item groups in GUIs

sgmedev:
[//]: # (The changelog will be added into CHANGELOG.md file automatically by fast-forward workflow.)

[//]: # (Please delete unrelevant changelog sections.)

### Feature Updates
- Added better naming for unknown (modded) item groups in GUIs

### Bug Fixes
- Fixed detection of unknown (modded) item groups in GUIs

### Others

[//]: # (The "Development Chores" section won't be included as release notes, it's recorded for developers only.)

### Development Chores

[//]: # (## Keep Writting If You Have Anything Else To Say)

### DIFF
--- a/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
+++ b/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
@@ -523,36 +523,37 @@ public class GroupGenerator {
     }
 
     /**
-     * Separate unknown slots into groups according to their position on the screen.
-     * Coordinates adjacency is used instead of slot.inventory() to calculate group belonging
-     * because some mods have: 1. non-adjacent slots in the same inventory 2. adjacent slots in different inventories.
+     * Separate unknown slots into groups according to their position on the screen,
+     * and give these group names according to slot instances' classes.
      *
      * @return Separation result
      */
     @NotNull
-    private static List<SlotsGroup> separateUnknownSlotsIntoGroups(List<SlotItem> unknownSlots) {
-        List<SlotsGroup> groupsOfUnknownSlots = new ArrayList<>();
-        List<List<SlotItem>> unknownGroups = new ArrayList<>(unknownSlots.size());
-        for (SlotItem slot : unknownSlots) {
-            unknownGroups.stream()
-                    .filter(group ->
-                            group.stream()
-                                    .anyMatch(groupSlot ->
-                                            groupSlot.x == slot.x && (groupSlot.y == slot.y + 18 || groupSlot.y == slot.y - 18)
-                                                    || groupSlot.y == slot.y && (groupSlot.x == slot.x + 18 || groupSlot.x == slot.x - 18)
-                                    )
-                    )
+    private static List<SlotsGroup> separateUnknownSlotsIntoGroups(List<SlotItem> slots) {
+        // Separate unknown slots into groups.
+        // Coordinates adjacency is used instead of slot.inventory() to calculate grouping
+        // because some mods have: 1. non-adjacent slots in the same inventory 2. adjacent slots in different inventories.
+        List<List<SlotItem>> separatedSlots = new ArrayList<>(slots.size());
+        for (SlotItem current : slots) {
+            // Search for a group which already has one slot that is adjacent to the current slot,
+            // or create a new group if there is no such group.
+            Optional<List<SlotItem>> targetGroup = separatedSlots.stream()
+                    .filter(group -> group.stream().anyMatch(groupSlot -> twoSlotsAreAdjacent(current, groupSlot)))
                     .findFirst()
                     .or(() -> {
-                        List<SlotItem> group = new ArrayList<>(unknownSlots.size());
-                        unknownGroups.add(group);
+                        List<SlotItem> group = new ArrayList<>(slots.size());
+                        separatedSlots.add(group);
                         return Optional.of(group);
-                    })
-                    .get()
-                    .add(slot);
+                    });
+            // Add the current slot to this group
+            targetGroup.ifPresent(group -> group.add(current));
         }
-        Map<String, Byte> usedNames = new HashMap<>(unknownGroups.size());
-        for (List<SlotItem> group : unknownGroups) {
+
+        List<SlotsGroup> result = new ArrayList<>();
+
+        // Naming groups
+        Map<String, Byte> usedNames = new HashMap<>(separatedSlots.size());
+        for (List<SlotItem> group : separatedSlots) {
             String groupName;
             boolean isOfSingularSlotType = group.stream()
                     .map(slot -> slot.slot.getClass())
@@ -575,10 +576,16 @@ public class GroupGenerator {
             } else {
                 usedNames.put(groupName, (byte) 1);
             }
-            groupsOfUnknownSlots.add(new SlotsGroup(groupName, group));
+            result.add(new SlotsGroup(groupName, group));
         }
 
-        return groupsOfUnknownSlots;
+        return result;
+    }
+
+    private static boolean twoSlotsAreAdjacent(SlotItem currSlot, SlotItem groupSlot) {
+        boolean adjacentOnX = groupSlot.x == currSlot.x && (groupSlot.y == currSlot.y + 18 || groupSlot.y == currSlot.y - 18);
+        boolean adjacentOnY = groupSlot.y == currSlot.y && (groupSlot.x == currSlot.x + 18 || groupSlot.x == currSlot.x - 18);
+        return adjacentOnX || adjacentOnY;
     }
 
     private static @NotNull List<SlotsGroup> inventoryAndCraftingScreensGroups(@NotNull HandledScreenAccessor screen) {
--- a/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
+++ b/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
@@ -489,23 +489,28 @@ public class GroupGenerator {
         }
         Map<String, Byte> usedNames = new HashMap<>(unknownGroups.size());
         for (List<SlotItem> group : unknownGroups) {
+            String groupName;
             if (group.stream()
                     .map(slot -> slot.slot.getClass().getSimpleName())
                     .distinct()
                     .limit(2)
                     .count() == 1) {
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
```diff
@@ -467,8 +466,51 @@ else if (screen.getHandler() instanceof HopperScreenHandler)
             foundGroups.add(itemOutputGroup);
         }
 
-        if (unknownGroup.slotItems.size() > 0) {
-            foundGroups.add(unknownGroup);
+        // Adjacency is used here instead of slot.inventory because some mods have non-adjacent slots in the same
```
boholder: It seems like these lines need indentation?

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
```diff
@@ -467,8 +466,51 @@ else if (screen.getHandler() instanceof HopperScreenHandler)
...
+        List<List<SlotItem>> unknownGroups = new ArrayList<>(unknownSlots.size());
+        for (SlotItem slot : unknownSlots) {
+            unknownGroups.stream()
+                    .filter(group ->
+                            group.stream()
+                                    .anyMatch(groupSlot ->
+                                            groupSlot.x == slot.x && (groupSlot.y == slot.y + 18 || groupSlot.y == slot.y - 18)
+                                            || groupSlot.y == slot.y && (groupSlot.x == slot.x + 18 || groupSlot.x == slot.x - 18)
+                                    )
+                    )
```
boholder: Can you extract these inner logic into methods?
Like:
 unknownGroups.stream()
                    .filter(GroupGenerator::<new method 1>)
                    .findFirst()
                    .or(GroupGenerator::<new method 2>)
                    .get()
                    .add(slot);

BTW I'm confused what this stream operation tries to do...

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
```diff
@@ -467,8 +466,51 @@ else if (screen.getHandler() instanceof HopperScreenHandler)
...
+                    .findFirst()
+                    .or(() -> {
+                        List<SlotItem> group = new ArrayList<>(unknownSlots.size());
+                        unknownGroups.add(group);
+                        return Optional.of(group);
+                    })
```
boholder, line 486: what's meaning of logic inside this or operation?
It seems like the result is always true since `new ArrayList<>(unknownSlots.size())` will never be null

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
```diff
@@ -467,8 +466,51 @@ else if (screen.getHandler() instanceof HopperScreenHandler)
...
+                    .add(slot);
+        }
+        Map<String, Byte> usedNames = new HashMap<>(unknownGroups.size());
+        for (List<SlotItem> group : unknownGroups) {
+            String groupName;
+            if (group.stream()
+                    .map(slot -> slot.slot.getClass().getSimpleName())
+                    .distinct()
+                    .limit(2)
+                    .count() == 1) {
```
boholder: this if condition deserves an independent bool value to explain its purpose

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
```diff
@@ -467,8 +466,51 @@ else if (screen.getHandler() instanceof HopperScreenHandler)
...
+                if (groupName.startsWith("class_")) {
+                    groupName = "unknown_group";
+                }
+            } else {
+                groupName = "unknown_group";
+            }
+            if (usedNames.containsKey(groupName)) {
+                byte n = (byte) (usedNames.get(groupName) + 1);
+                usedNames.put(groupName, n);
+                groupName += String.format("_%d", n);
```
boholder: Really nice idea about getting slot group names.
It misses concerns about I18N but I can't come up with better idea to amend it.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
```diff
@@ -467,8 +466,51 @@ else if (screen.getHandler() instanceof HopperScreenHandler)
...
+            } else {
+                usedNames.put(groupName, (byte) 1);
+            }
+            foundGroups.add(new SlotsGroup(groupName, group));
```
boholder: What's the value of groupName when code goes to here,
for example, number1_number2_number3... ?
This is not readable than unknown group in my view, maybe I misunderstood the meaning of these group naming logic.
And maybe extract this part (in for loop) as a method is good.

boholder (changes requested):
Forgive my stupid, I can't understand some details in the PR, can you make the code more clear?

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
sgmedev: Some kind of I18N would be a good idea though so that it doesn't say `minecraft_access.slot_group.<name> group` even if it's not proper I18N and it just says the (usually) English name, but I dont have enough experience with I18N to implement such a thing.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
sgmedev: Essentially, if all the slots in a group have the same class name groupName will be set to that name, if groupName starts with `class_` (vanilla obfuscated classes) or the slots have different classes it will be set to unknown_group. Then if there is already a group with that name it will get a number added to the end.

This approch does have the side effect of not giving the first group a number when there are other groups with the same name but the alternatives are giving every group a number or making the algorithm significantly more complex.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
sgmedev: The outer stream operation looks for an unknown group which already has adjacent slots in it using the inner stream which checks if any slots in the group are adjacent, if it doesn't find any it creates a new group and returns that instead.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
sgmedev: I was using neovim to make some quick edits this morning and for whatever ungodly reason it insisted on using tabs for indentation, I though I got all of them but there are probably a few lines that I didn't catch.

sgmedev: Those particular lines do look indented properly though and I made them in idea, I've fixed the one line that Neovim nicely stuffed for me.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
sgmedev: Upon further inspection it doesn't look like it's practical to extract those to methods because they reference stuff in the outer scope so I wouldn't be able to use method references.

sgmedev:
I've moved that condition to a boolean and fixed the one indentation error I could see. Short of adding a bunch of comments I don't think there is much I can do to make it clearer but I might be able to have a better look and think of something if I have another look when it's not late afternoon after a party.

boholder:
> Short of adding a bunch of comments I don't think there is much I can do to make it clearer

Use method naming, variable naming to describe the purpose of logic, instead of comments.
Only comment when you can't find a brief clear conclusion about the logic.

See https://github.com/FabricMC/yarn/blob/24w03b/CONVENTIONS.md

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
boholder: Yeah, I'm wrong, sorry.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
boholder: Now it make sense to me. Can you add some comments about describing these if-else conditions like what you did in your last comment?

boholder: TIL: use byte for small numbers. I don't know if it's a good usage...

boholder:
Some refactor FYI: https://github.com/boholder/minecraft-access/commits/modded-gui-groups/

The or() can be replaced with orElse() thus the result can be directly List without Optional wrapper

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/inventory_controls/GroupGenerator.java
sgmedev: > TIL: use byte for small numbers. I don't know if it's a good usage...

There would have to be 128 identically named groups for it to be a problem, I could use a short if that's a concern but I have never seen a mod that has anywhere near 128 slot groups that I can remember and it should wrap around to -128 if that does happen for some reason, realistically the detection of 128 groups is probably itself a bug or a really bad ui and a better mekanism for switching groups may be in order for that many groups. I guess we're not going to be ram limited to the point that a few bytes is significant and it should theoretically have roughly O(n) space complexity so there is no reason not to use a short or even an int or a long.

> Now it make sense to me. Can you add some comments about describing these if-else conditions like what you did in your last comment?

I shall look at that tomorrow.

boholder (approved):
LGTM 👍

khanshoaib3:
\[Fast Forward CI\] modded-gui-groups merged into 1.20... Copied the changelogs from pull request body to doc/CHANGELOG.md... Pushed the changes to origin.

PR #301, opened 2024-07-22
fix(forge): it seems like gradle build will break when forge module doesn't have remapJar task

boholder:
:p


khanshoaib3 (approved):

khanshoaib3:
\[Fast Forward CI\] 1.20-fix merged into 1.20...\nNo changelogs found... Pushed the changes to origin.

PR #333, opened 2024-08-30
Add information about using a screenreader

douglassquirrel:
- Updated modpack setup instructions to describe when users might want or need a screenreader such as NVDA or JAWS.


khanshoaib3:
\[Fast Forward CI\] patch-1 merged into 1.21... Copied the changelogs from pull request body to docs/changelog/latest.md... Pushed the changes to origin.

PR #334, opened 2024-08-30
Add Jade an an alternative read-crosshair backend

### LINKED ISSUE #325: Support Jade as an alternative block narration backend
[Jade](https://modrinth.com/mod/jade) shows information about the currently targeted block including container contents and I18N and has a binding for the narrator to read this information aloud.

Currently to use Jade's TTS you have to either press the TTS button to read once or set the TTS button to be a toggle in the config  to read automatically however this currently conflicts with MC Access and prevents using the TTS button to read on demand. Using the toggle also makes it easy to accidentally disable leaving the user with nothing.

A Jade integration would have MC Access call Jade's TTS method if Jade is detected instead of using our own implementation providing a much more detailed description.

A config option may be desired for users who like the current terse reading but want to use Jade's TTS key for on-demand details

Jade's TTS method:
https://github.com/Snownee/Jade/blob/380842dd425f3e252b51dbe52070b7889725ac84/src/main/java/snownee/jade/overlay/WailaTickHandler.java#L58-L76

And where it is called when using the TTS key:
https://github.com/Snownee/Jade/blob/380842dd425f3e252b51dbe52070b7889725ac84/src/main/java/snownee/jade/JadeClient.java#L150

CC: @braillebennett

sgmedev:
Fixes #325 

# Changelog
## New Features
- Blocks and entities can now be narrated with Jade, providing much more detailed information.


common/src/main/java/com/github/khanshoaib3/minecraft_access/features/read_crosshair/CrosshairNarrator.java
```diff
@@ -0,0 +1,31 @@
...
+     * Determines weather the player is looking at the same block or entity.
+     * {@link #narrate(ClientWorld, boolean)} will only be called
+     * if this method returns a different non-null value than its previous call.
+     * A value of null indicates that the player is not currently looking at any block
+     * and thus {@link #narrate(ClientWorld, boolean)} should not be called.
+     * @param world The current world the player is in
+     * @param speakSide Weather the block side the player is pointing at should be narrated
+     * @return Any object comparable with {@link java.util.Objects#equals(Object, Object)} or null
+     */
+    @Nullable Object deduplication(@NotNull ClientWorld world, boolean speakSide, boolean speakConsecutiveBlocks);
```
boholder: `deduplication` is not a clear name to explain its function, how about `getPlayerLookingAt`.
the `ClientWorld` param can be replaced by internal `WorldUtils.getClientWorld`.
(And... detailed method doc is good, but the explanation about Crosshair speaking isn't necessary here, this part can be moved to invocation of this method)

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/read_crosshair/CrosshairNarrator.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Generates a description of the block or entity the player is currently looking at to be read out.
+     * This will never be called if {@link #deduplication(ClientWorld, boolean, boolean)} returns null.
+     * @param world The current world the player is in
+     * @param speakSide Weather the block side the player is pointing at should be narrated
+     * @return A non-null string to be read to the player
+     */
+    @NotNull String narrate(@NotNull ClientWorld world, boolean speakSide);
```
boholder: `narrate`'s param can be the return value of `deduplication`, then we don't need `rayCast`.

sgmedev: `rayCast` exists because I noticed that Jade's ray casting was subtly different and I wanted everything to line up with Jade if it's active. I don't think passing `deduplication`'s return value to `narrate` would be very useful and it would be a bit of a mess with generics because `deduplicatiom`'s return type is currently `Object`. All methods in the `CrosshairNarrator` interface are completely different between both implementations so i don't think it makes sense to move anything to `NarratorUtils`.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/read_crosshair/CrosshairNarrator.java
sgmedev: The method's sole purpose is to indicate when the target changes and prevent unwanted duplicated narration (deduplication), it's equivalent to the old `query` variable. It only returns an Object that can be compared for changes, implementers can make it return anything that when changed indicates that the target has changed.

I just included the `world` param for convenience because I wasn't aware of `WorldUtils.getClientWorld` and didn't want to go through all the null checks neccessary to get it otherwise, I'll look into changing that in the morning.

I feel it's important to include the relationship between `deduplication` and `narrate` so that implementers (potentially in a future API) understand how it works without having to dig around the callsite.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/read_crosshair/CrosshairNarrator.java
```diff
@@ -0,0 +1,31 @@
...
+    @NotNull HitResult rayCast();
+
+    /**
+     * Determines weather the player is looking at the same block or entity.
+     * {@link #narrate(ClientWorld, boolean)} will only be called
+     * if this method returns a different non-null value than its previous call.
+     * A value of null indicates that the player is not currently looking at any block
+     * and thus {@link #narrate(ClientWorld, boolean)} should not be called.
+     * @param world The current world the player is in
+     * @param speakSide Weather the block side the player is pointing at should be narrated
```
boholder: Typo: Weather  -> Whether (might be made by me)

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/read_crosshair/ReadCrosshair.java
```diff
@@ -1,28 +1,22 @@
...
 import net.minecraft.sound.SoundEvents;
 import net.minecraft.util.Identifier;
-import net.minecraft.util.Pair;
 import net.minecraft.util.hit.BlockHitResult;
 import net.minecraft.util.hit.EntityHitResult;
 import net.minecraft.util.hit.HitResult;
-import net.minecraft.util.math.BlockPos;
-import net.minecraft.util.math.Direction;
 import net.minecraft.util.math.Vec3d;
+import org.jetbrains.annotations.Nullable;
```
boholder, line 19: I still don't know whether jetbrains's annotations will affect vscode users

sgmedev: They're ubiquitous enough that I'd hope they would, worst case they'll simply be ignored by vscode and have no effect.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/read_crosshair/CrosshairNarrator.java
boholder: I know your purpose, but it doesn't fit current function signature: world, speakSide in, object out... Another choice is pass previous object into this method, then make it return a boolean value to indicate whether to speak.
ACK about other comments.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/read_crosshair/CrosshairNarrator.java
sgmedev: Oopsie, ill fix that in the morning. BTW, when you make a PR review comment there is a button above the comment box to make a suggestion which I can then accept with a simple push of a button. I'm surprised Grazie didn't pick tat up.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/read_crosshair/ReadCrosshair.java
boholder: > they'll simply be ignored by vscode and have no effect.

Without a "class not found" hint? I'm not sure...

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/read_crosshair/CrosshairNarrator.java
boholder: > there is a button above the comment box to make a suggestion which I can then accept with a simple push of a button.

Ok, I'll try to find it in Braille's PR... let me read the GitHub doc...

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/read_crosshair/CrosshairNarrator.java
sgmedev: I've removed the `world` parameter, turns out I wasn't using it anymore anyway. I don't quite understand what you mean about the signature not matching the name, I think `deduplication` aptly describes the purpose of the method and is the most appropriate name.

khanshoaib3:
\[Fast Forward CI\] jade merged into 1.21... Copied the changelogs from pull request body to docs/changelog/latest.md... Pushed the changes to origin.

PR #346, opened 2024-09-29
Fix mod crashing when running server-side

Molitvan:
- Removed unnecessary use of Fabric api in a few random places
- Deleted MainClassFabric because the mod does two completely different things depending on which side it's running on and a common entry point for both sides is not needed (and also unnecessarily complicates things with the server-side error message)

# Changelog
## Others
- Added an error message when trying to run the mod server-side


khanshoaib3:
\[Fast Forward CI\] server-side-not-supported-message merged into 1.21... Copied the changelogs from pull request body to docs/changelog/latest.md... Pushed the changes to origin.

PR #349, opened 2024-10-05
Access menu shortcuts bar and some refactor

boholder:
with https://github.com/khanshoaib3/minecraft-access-i18n/pull/47

# Changelog
## Feature Updates
- **BREAKING CHANGES**: Rename **Narrator Menu** aka F4 Menu to Access Menu
- **Access Menu**: Access Menu functions can be triggered with Alt + upper number keys


boholder:
Test failed, gonna fix it

common/src/main/java/com/github/khanshoaib3/minecraft_access/mixin/MinecraftClientMixin.java
```diff
@@ -0,0 +1,26 @@
...
+public class MinecraftClientMixin {
+    /**
+     * {@link AccessMenu} allows menu functions to be triggered when
+     * no screen opened and alt key with number key are pressed.
+     * We need to suppress original hotbar slot selecting feature.
+     */
+    @Inject(at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;isSpectator()Z"),
+            method = "handleInputEvents",
+            cancellable = true)
+    private void suppressHotbarSlotSelectingWhenAccessMenuFunctionsWork(CallbackInfo ci) {
```
sgmedev: This method name is way too long, consider naming it `handleInputEvents` to match where it is being injected into.
```suggestion
    private void handleInputEvents(CallbackInfo ci) {
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/system/KeyUtils.java
```diff
@@ -93,6 +93,10 @@ public static boolean isRightAltPressed() {
         return isAnyPressed(GLFW.GLFW_KEY_RIGHT_ALT);
     }
 
+    public static boolean isLeftOrRightAltPressed() {
```
sgmedev: This method name is a bit verbose, `isAltPressed` would be better.
```suggestion
    public static boolean isAltPressed() {
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/condition/TimingKeystroke.java
```diff
@@ -9,22 +9,22 @@
  * For checking keystroke conditions that related to time,
  * like "interval/rate-limitation-on-feature-execution", "double/triple-click"
  */
-public abstract class KeystrokeTiming extends Keystroke {
+public abstract class TimingKeystroke extends Keystroke {
```
sgmedev: Consider `TimedKeystroke`.
```suggestion
public abstract class TimedKeystroke extends Keystroke {
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/condition/MenuKeystroke.java
```diff
@@ -24,6 +24,9 @@ public MenuKeystroke(BooleanSupplier condition) {
         super(condition);
     }
 
+    /**
+     * Pure function
+     */
```
sgmedev: Consider using JetBrains' `@Contract` annotation to get warnings for incorrect usage.
```suggestion
    @Contract(pure = true)
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/condition/Keystroke.java
```diff
@@ -9,6 +14,9 @@
  * For keys that only execute once until released and pressed again.
  */
 public class Keystroke {
+
+    protected static List<Keystroke> EXISTING_KEYSTROKES = new ArrayList<>();
```
sgmedev: This is not a constant and thus should be in camelCase, not SCREAMING_SNAKE_CASE. `keystrokes` may also be a better name because "existing" doesn't really add any meaning. This field can also be final.
```suggestion
    protected static final List<Keystroke> keystrokes = new ArrayList<>();
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/condition/Keystroke.java
```diff
@@ -131,4 +147,12 @@ public boolean aboutToHappen(Keystroke keystroke) {
             return this.preCondition.apply(keystroke);
         }
     }
+
+    /**
+     * Let this method handle state updates,
+     * no need to manually call the update method in every feature.
+     */
+    public static void updateAllExistingKeyStrokesStates() {
```
sgmedev: Something like `updateKeystrokes` would be more clear.
```suggestion
    public static void updateKeystrokes() {
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/MainClass.java
```diff
@@ -3,12 +3,15 @@
 import com.github.khanshoaib3.minecraft_access.config.Config;
 import com.github.khanshoaib3.minecraft_access.config.config_maps.*;
 import com.github.khanshoaib3.minecraft_access.features.*;
-import com.github.khanshoaib3.minecraft_access.features.inventory_controls.InventoryControls;
 import com.github.khanshoaib3.minecraft_access.features.access_menu.AccessMenu;
+import com.github.khanshoaib3.minecraft_access.features.HealthAndHunger;
```
sgmedev: Redundant import, included in wildcard on line 5:
```java
import com.github.khanshoaib3.minecraft_access.features.*;
```
```suggestion
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/config/ConfigMap.java
```diff
@@ -28,8 +28,8 @@ public class ConfigMap {
     private FallDetectorConfigMap fallDetectorConfigMap;
     @SerializedName("Read Crosshair")
     private ReadCrosshairConfigMap readCrosshairConfigMap;
-    @SerializedName("Narrator Menu")
-    private NarratorMenuConfigMap narratorMenuConfigMap;
+    @SerializedName("Access Menu")
```
sgmedev: This is a breaking change and should be documented as such.

common/src/main/java/com/github/khanshoaib3/minecraft_access/MainClass.java
```diff
@@ -124,24 +127,27 @@ private static void _clientTickEventsMethod(MinecraftClient minecraftClient) {
 
         PositionNarrator.getInstance().update();
 
-        if (healthNHunger != null && otherConfigsMap.isHealthNHungerEnabled())
-            healthNHunger.update();
+        boolean worldAndPlayerExists = Objects.nonNull(MinecraftClient.getInstance()) && Objects.nonNull(WorldUtils.getClientPlayer());
+        if (worldAndPlayerExists) {
```
sgmedev: `MinecraftClient.getInstance()` is never null, and thus checking for it is superfluous. This is also significantly clearer as a regular null check as opposed to using a variable.
```suggestion
        if (WorldUtils.getClientPlayer() != null) {
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/MouseKeySimulation.java
```diff
@@ -23,23 +20,14 @@
  * 4) mouse wheel scroll up<br>
  * 5) mouse wheel scroll down
  */
-@Slf4j
 public class MouseKeySimulation {
-    private static final MouseKeySimulation instance;
-
-    private boolean enabled;
+    private static boolean ENABLED;
```
sgmedev: This is not a constant.
```suggestion
    private static boolean enabled;
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/HealthAndHunger.java
```diff
@@ -0,0 +1,31 @@
...
+import com.github.khanshoaib3.minecraft_access.utils.condition.Keystroke;
+import lombok.extern.slf4j.Slf4j;
+import net.minecraft.client.resource.language.I18n;
+
+/**
+ * Speak Player Health and Hunger Key (default: R)
+ */
+@Slf4j
+public class HealthAndHunger {
+    private static final Keystroke key = new Keystroke(KeyBindingsHandler.getInstance().healthNHungerNarrationKey);
```
sgmedev: Should the key be renamed to be consistent with everything else?
```suggestion
    private static final Keystroke key = new Keystroke(KeyBindingsHandler.getInstance().healthAndHungerNarrationKey);
```

docs/features.md
```diff
@@ -209,9 +209,9 @@ See also: [Configuration](/docs/config.md#position-narrator), [Keybindings](/doc
 
 ### Health n Hunger
```
sgmedev: Should this be renamed to be consistent with the name used elsewhere?
```suggestion
### Health and Hunger
```

docs/keybindings.md
```diff
@@ -127,20 +127,26 @@ See also: [Feature Description](/docs/features.md#position-narrator), [Configura
 
 ## Health n Hunger
```
sgmedev: Should this be renamed to be consistent with the name used elsewhere?
```suggestion
## Health and Hunger
```

sgmedev (changes requested):
Looks good, just a few minor considerations mainly with naming and a lack of consistency with whether it's named `Health N Hunger` or `Health and Hunger`, I personally prefer the latter. The renaming of `Narrator Menu` config option to `Access Menu` should be included in the changelog as it is a breaking change.

common/src/main/java/com/github/khanshoaib3/minecraft_access/MainClass.java
boholder: I think if this mod is running on server, that instance is null.

boholder:
@TheSuperGamer20578 perfect review! It exposures many naming issues. Thank you for pointing out them!

common/src/main/java/com/github/khanshoaib3/minecraft_access/MainClass.java
sgmedev: That would make sense, I kind of only looked at the constructor for `MinecraftClient` and saw that it was always set there forgetting that the constructor is only called from the client. I still feel that the additional nesting that `Objects.nonNull()` adds makes it significantly harder to read, the [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#nonNull-java.lang.Object-) for it states that it exists for use as a predicate for methods such as `Stream.filter()`. Using a variable for this condition is also superfluous because a regular null check is usually easy enough to read.
```suggestion
        if (MinecraftClient.getInstance() != null && WorldUtils.getClientPlayer() != null) {
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/condition/Interval.java
```diff
@@ -56,12 +44,12 @@ public boolean isReady() {
      *
      * @return true if the delay timer has stopped or cooled down.
      */
-    public boolean hasEnded() {
-        if (!this.isRunning) return true;
-        if (!this.isReady()) return false;
+    public boolean hasNotEnded() {
```
sgmedev: Having a not in a predicate name can be confusing, especially if the result is ever inverted as that would result in a double negative (`!hasNotEnded()`). Something like `isRunning` would be more clear; however, I see that there is already a field with that name. Additionally, the Javadoc states that it will not auto-reset the timer but then `isReady` is called which will auto-reset the timer.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/MouseKeySimulation.java
```diff
@@ -49,15 +49,16 @@ public class MouseKeySimulation {
...
 
     private static void loadConfigurations() {
         MouseSimulationConfigMap map = MouseSimulationConfigMap.getInstance();
-        ENABLED = map.isEnabled();
-        mouseScrolls[0].setInterval(Interval.inMilliseconds(map.getScrollDelayInMilliseconds(), mouseScrolls[0].interval()));
-        mouseScrolls[1].setInterval(Interval.inMilliseconds(map.getScrollDelayInMilliseconds(), mouseScrolls[1].interval()));
+        enabled = map.isEnabled();
+        Interval scrollInterval = Interval.ms(map.getScrollDelayInMilliseconds());
+        mouseScrolls[0].interval = scrollInterval;
+        mouseScrolls[1].interval = scrollInterval;
```
sgmedev: Should these be different intervals?

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/point_of_interest/LockingHandler.java
```diff
@@ -161,8 +160,7 @@ private void mainLogic() {
                     PlayerUtils.playSoundOnPlayer(SoundEvents.BLOCK_NOTE_BLOCK_PLING, aimAssistAudioCuesVolume, bowState);
                     lastAimAssistCue = 1;
                 }
-            }
-            else {
+            } else {
                 if (lastAimAssistCue != 0 || bowState != lastBowState) {
```
sgmedev: Should this not be an else if statement?
```suggestion
            } else if (lastAimAssistCue != 0 || bowState != lastBowState) {
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/condition/Keystroke.java
```diff
@@ -152,7 +152,7 @@ public boolean aboutToHappen(Keystroke keystroke) {
      * Let this method handle state updates,
      * no need to manually call the update method in every feature.
      */
-    public static void updateAllExistingKeyStrokesStates() {
-        EXISTING_KEYSTROKES.forEach(Keystroke::updateStateForNextTick);
+    public static void updateAllInstantsState() {
```
sgmedev: Typo: `Instants` → `Instances`

I also feel that something along the lines of `updateInstances` or `updateAllStates` would be better because `updateAllInstancesState` is a bit verbose.

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/system/KeyUtils.java
```diff
@@ -50,31 +44,21 @@ public static boolean isAllPressed(int... keyCodes) {
...
-                        // If this keybinding is bound to a keyboard-key,
-                        // let's use our key-pressing-check logic to circumvent the limitation.
-                        return KeyUtils.isOnePressed(keyCode);
-                    } else {
-                        // If this keybinding is bound to a non-keyboard key, execute the original method.
-                        return b.isPressed();
-                    }
-                }
-        );
+        return Arrays.stream(keyBindings).anyMatch(isKeyPressed());
```
sgmedev: A method reference may be better than returning a lambda.
```suggestion
        return Arrays.stream(keyBindings).anyMatch(this::isKeyPressed);
```

sgmedev (changes requested):
Just a few more minor tweaks and then I think it should be ready to merge.

sgmedev:
I’ve also discovered that the new shortcuts trigger on every tick while held and `alt+4` and `alt+5` cause the game to freeze for a good few seconds, which is exacerbated by them triggering many times when you hold them down.

boholder:
> I’ve also discovered that the new shortcuts trigger on every tick while held and `alt+4` and `alt+5` cause the game to freeze for a good few seconds, which is exacerbated by them triggering many times when you hold them down.

Roger, ~~I'll fix it and other refactors in Sunday.~~

20241013 edit: My head isn't clear... I want to delay the fixing for another week.

common/src/main/java/com/github/khanshoaib3/minecraft_access/MainClass.java
boholder: Yep, you're right, will accept it.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/MouseKeySimulation.java
boholder: ~~They are different IntervalKeystroke instances, so...~~ oh I get you

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/point_of_interest/LockingHandler.java
boholder: This class is complicate, I already refactored it but it's still complicated, @Molitvan 's bow aiming assist added more complexity. That's why I want to recode the POI and locking system.

boholder:
@TheSuperGamer20578 I suggest you review the new part commit by commit, I followed the "small commit" principal.

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/condition/Keystroke.java
sgmedev: How has this been resolved? I don’t see any changes

boholder: > How has this been resolved? I don’t see any changes

@TheSuperGamer20578 Now the code is:
https://github.com/khanshoaib3/minecraft-access/pull/349/files#diff-adafc93f481283c43ffb98839e6fca7f44e2ef8e069104f7544b781ffc0e66ccL133
```
    /**
     * Let this method handle state updates,
     * no need to manually call the update method in every feature.
     */
    public static void updateAllInstantsState() {
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/access_menu/AccessMenu.java
```diff
@@ -38,71 +37,53 @@ public class AccessMenu {
...
+        // two long-time-running find-the-closest-liquid-source functions need a long interval (10 seconds)
+        functionIntervals[4] = Interval.ms(10_000);
+        functionIntervals[5] = Interval.ms(10_000);
+
+        for (int i = 0; i < 10; i++) {
+            FUNCTIONS[i].keystroke.interval = functionIntervals[i];
+        }
+    }
+
+    private record MenuFunction(int number, IntervalKeystroke keystroke, Runnable func) {
```
sgmedev: It would be better to add the interval here instead of using an index and separate array; along with alleviating the need to keep multiple arrays in sync, this also eliminates the need for magic numbers as seen on lines +78 and +79. Though I feel that this should be completely refactored so that it only needs to use a single array as opposed to two arrays in seperate classes that must be kept in sync.

common/src/main/java/com/github/khanshoaib3/minecraft_access/mixin/BookEditScreenMixin.java
```diff
@@ -88,19 +88,16 @@ public void render(DrawContext context, int mouseX, int mouseY, float delta, Cal
         if (minecraft_access$tabKey.isPressed()) {
             minecraft_access$switchMouseHoveredButton();
         }
-        // update the state
-        minecraft_access$tabKey.updateStateForNextTick();
 
         if (minecraft_access$spaceKey.isPressed()) {
-            minecraft_access$spaceKey.updateStateForNextTick();
+
```
sgmedev: Is there supposed to be a blank line here?
```suggestion
```

sgmedev (changes requested):
`IntervalKeystroke` needs some work. My testing immediately found that camera controls are quite flakey as to whether or not they work when keys are tapped and that sometimes double taps are ignored, though the double tap behaviour is also observed on the `1.21` branch.

What should happen is that on key down is:

1. The action should always be executed promptly and unconditionally

2. The interval should be reset to prevent inconsistent timing of the first repetition

When the user taps a key, it is expected that the action will be executed promptly once, if the user double-taps a key they expect the action to be executed promptly a second time.

The current, undesirable behaviour can be observed in this screen recording:

https://github.com/user-attachments/assets/e7674f88-0ae9-49fc-b95e-a51546ccc304

I also found some issues with the access menu implementation; however, this can wait for a later refactor.

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/condition/Keystroke.java
sgmedev: Is that not the same as the code I reviewed earlier and suggested a change on?

common/src/main/java/com/github/khanshoaib3/minecraft_access/mixin/BookEditScreenMixin.java
boholder: Ah... ignore it, IDEA won't auto format it, won't fix.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/access_menu/AccessMenu.java
boholder: I guess we can fix it with #364  won't fix in this PR.

boholder:
> `IntervalKeystroke` needs some work.

Will look at it this weekend. And... if I understand clearly, this is an old bug, not produced by this PR, how about start another issue for it and won't fix it in release 1.9.0?

common/src/main/java/com/github/khanshoaib3/minecraft_access/utils/condition/Keystroke.java
boholder: > Is that not the same as the code I reviewed earlier and suggested a change on?

I don't know...

boholder:
> `IntervalKeystroke` needs some work.

The real reason for this issue is the class scope "global" `Interval` instances in `CameraControls`, `InventoryControls`, `LockingHandler` that manually triggered by player's keystrokes while having an interval limitation, the interval isn't properly working with keystrokes.
Meanwhile the totally auto features like `POIBlock` won't be affected.
Shoaib's previous fix linked interval and keystrokes, ~~I'll simple replace all the keystrokes to `IntervalKeystroke`~~ it seems that his solution is simplest one.

common/src/main/java/com/github/khanshoaib3/minecraft_access/config/config_menus/NarratorMenuConfigMenu.java
BrailleBennett, line 1: This file should be access menu not narrator menu

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/HealthAndHunger.java
BrailleBennett, line 1: Can you roll back health and hunger and related i18n changes? PR #371 adds your changes and more as discussed.

BrailleBennett (changes requested):

sgmedev (dismissed):
LGTM! I noticed that the `IntervalKeystroke` behaviour still isn't optimal, however there is no regression and this can be fixed later.

boholder:
> I noticed that the `IntervalKeystroke` behaviour still isn't optimal

I made the change and reverted it, since there is no existing example for me to test the behavior... not hard, just add relative logic in `updateForNextTick` method.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/HealthAndHunger.java
boholder: @BrailleBennett Sorry, I can't, it's burred under commits, can't simply revert it.

common/src/main/java/com/github/khanshoaib3/minecraft_access/config/config_menus/NarratorMenuConfigMenu.java
boholder: @BrailleBennett I'm afraid this will be resolved in your PR?...

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/HealthAndHunger.java
sgmedev: You can use `git revert` to revert a commit, this basically just creates a new commit that undoes any changes the specified commit made, otherwise you can merge it into here or wait until it gets merged and then resolve the merge conflicts which will likely effectively revert the changes

CONTRIBUTING.md
```diff
@@ -118,7 +117,7 @@ is for providing user information to client side game when developing.
         * `minecraft_access.accesswidener`: Yet another configuration.
         * `minecraft_access-common.mixins.json`: Mixin configuration that tells the Mixin framework where to find Mixin classes under `common` directory.
         * `assets/minecraft_access`: Contains custom resource files of this mod, structured in the [Minecraft Resource Pack](https://minecraft.wiki/w/Resource_pack) format.
-            * `lang`: The I18N text used by this mod for narration purpose, managed with [Weblate].
+            * `lang`: The I18N text used by this mod for narration purpose, managed with [Weblate](https://hosted.weblate.org/engage/minecraft-access/).
```
sgmedev: This is not necessary, the URL is already at the end of the file

sgmedev (changes requested):

CONTRIBUTING.md
boholder: But, it's quoted with [ ]...

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/HealthAndHunger.java
boholder: So... please refer https://github.com/minecraft-access/minecraft-access/pull/349/commits/e557170467c4f1c26670c3525eace5a70e574bbe for modifying doc, refer https://github.com/minecraft-access/minecraft-access/pull/349/files#diff-9c439a76705ddf2f13f81ed1853e75bfd077a5a2163db1e34bf5f3d3b816d3dcR126 for modifying triggering logic, refer https://github.com/minecraft-access/minecraft-access/pull/349/commits/559c27dda31a05ee0ab9e19f3726ebf259ced313 for utilizing the functions, in #371 
I'll revert my changes.

CONTRIBUTING.md
sgmedev: In markdown you can use `[text]` and then put `[text]: url` at the end of the file, this is equivalent to `[text](url)`. I generally prefer this syntax because it makes lines significantly shorter and thus greatly inproves readability.

boholder: TIL!

sgmedev (dismissed):

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/HealthAndHunger.java
boholder: @boholder TODO

common/src/main/java/com/github/khanshoaib3/minecraft_access/config/config_menus/NarratorMenuConfigMenu.java
boholder: @boholder TODO

BrailleBennett:
Make sure to update the changelog to reflect the new additions this PR does.

boholder:
Holy...

BrailleBennett (approved):

PR #379, opened 2024-11-06
Fix entity equipment narration triggering even if entity doesn't have equipment

### LINKED ISSUE #368: Narration erroneously says the word "with" after entity names when the entity isn't holding equipment
When an entity is being narrated by the built in Minecraft access narration, there is no check to see if the word "with" should be included after entity names. Usually this makes sense when an entity is holding something (E.G.: Skeleton with bow), but when the entity isn't holding anything then this word doesn't make any sense to be appended after the entity's name (E.G. Pig with) should just be pig.

Molitvan:
Fixes #368 

# Changelog
## Bug Fixes
- Fix equipment narration triggering even if entity doesn't have equipment


khanshoaib3 (approved):

sgmedev (approved):
LGTM!

PR #392, opened 2024-11-16
Fix overlapping narration on load

### LINKED ISSUE #352: Speech overridden when loading into a world
When loading into a world both the current biome and the playerr's xp level are queued, the current biome message is set as interruptible, so the player's vcurrent xp level immediantely overwrites it and the biome is never heard.

BrailleBennett:
# Changelog
## Feature Updates
- The mod now announces the biome you entered when you join a world instead of the xp level.

### DIFF
--- a/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/BiomeIndicator.java
+++ b/common/src/main/java/com/github/khanshoaib3/minecraft_access/features/BiomeIndicator.java
@@ -6,45 +6,35 @@ import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.resource.language.I18n;
 import net.minecraft.registry.entry.RegistryEntry;
 import net.minecraft.world.biome.Biome;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Narrates the name of the biome when entering a different biome.
  */
 @Slf4j
 public class BiomeIndicator {
-    private String previousBiome;
+    @Nullable
+    private String previousBiome = null;
 
     public BiomeIndicator() {
-        previousBiome = "";
     }
 
     public void update() {
-        try {
-            MinecraftClient minecraftClient = MinecraftClient.getInstance();
-            if (minecraftClient == null) return;
-            if (minecraftClient.world == null) return;
-            if (minecraftClient.player == null) return;
-            if (minecraftClient.currentScreen != null) return;
+        MinecraftClient minecraftClient = MinecraftClient.getInstance();
+        if (minecraftClient == null) return;
+        if (minecraftClient.world == null) return;
+        if (minecraftClient.player == null) return;
+        if (minecraftClient.currentScreen != null) return;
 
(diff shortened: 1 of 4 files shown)


boholder (dismissed):
LGTM

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/BiomeIndicator.java
```diff
@@ -6,45 +6,35 @@
...
  */
 @Slf4j
 public class BiomeIndicator {
-    private String previousBiome;
+    @Nullable
+    private String previousBiome = null;
 
     public BiomeIndicator() {
-        previousBiome = "";
     }
```
sgmedev: This can be removed entirely. Java will create a public no args constructor by default if there is none.

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/BiomeIndicator.java
```diff
@@ -6,45 +6,35 @@
...
+        RegistryEntry<Biome> var27 = minecraftClient.world.getBiome(minecraftClient.player.getBlockPos());
+        String name = I18n.translate(getBiomeName(var27));
+
+        if (previousBiome == null) {
+            previousBiome = name;
+            MainClass.speakWithNarrator(I18n.translate("minecraft_access.other.biome", name), true);
+        } else if (!previousBiome.equalsIgnoreCase(name)) {
+            previousBiome = name;
+            MainClass.speakWithNarrator(I18n.translate("minecraft_access.other.biome", name), true);
         }
```
sgmedev: Duplicated code. String.equalsIgnoreCase() will also return false if its argument is null so an explicit null check is superfluous.
```suggestion
        if (!name.equalsIgnoreCase(previousBiome)) {
            previousBiome = name;
            MainClass.speakWithNarrator(I18n.translate("minecraft_access.other.biome", name), true);
        }
```

common/src/main/java/com/github/khanshoaib3/minecraft_access/features/XPIndicator.java
```diff
@@ -5,37 +5,39 @@
...
+        if (previousXPLevel == null) {
             previousXPLevel = currentXPLevel;
-
-            String toSpeak = (increased) ? I18n.translate("minecraft_access.xp_indicator.increased", currentXPLevel)
-                    : I18n.translate("minecraft_access.xp_indicator.decreased", currentXPLevel);
-            MainClass.speakWithNarrator(toSpeak, true);
-        } catch (Exception e) {
-            log.error("An error occurred in xp indicator.", e);
+            return;
+        } else if (previousXPLevel == currentXPLevel) {
```
sgmedev: Using an else if here is redundant because there is already a return statement in the first branch.
```suggestion
        }
        if (previousXPLevel == currentXPLevel) {
```

sgmedev (changes requested):

sgmedev (approved):

PR #347, opened 2024-10-02
Switch to Cloth Config

sgmedev:
Closes #118 

# Changelog

## New Features

- Support navigating and narrating on the config menu of the [Cloth Config](https://shedaniel.gitbook.io/cloth-config) library, this provides support to the config menus of other mods that also use this library. (If you run into a mod's config menu that is built with Cloth that isn't accessible, report it as a bug)
 
## Feature Updates

- Cloth Config is now used for this mod's config menu, the location of the config menu is unchanged however, and remains in the `Access Menu`.
- New `Features` config category that contains on-off switches for some automatic features.

PR #397, opened 2024-11-30
Removed some copied i18n strings and replaced them with vanilla versions

BrailleBennett:
# Changelog
## Translation Changes
- Removed coppied strings that were already in use in vanilla, mainly the seperate color strings that Minecraft access had which vanilla translation already provides.


BrailleBennett:
Only known issue is entity narration broke and I have no idea why. I didn't touch that part of i18n at all as you can see in the diff. I also didn't touch config keys, as #347 addresses them indirectly as cloth config uses vanilla keys where applicable already.

common/src/main/resources/assets/minecraft_access/lang/en_us.json
boholder, line 1: Can you also remove these items in other language files?

boholder (changes requested):
LGTM

common/src/main/resources/assets/minecraft_access/lang/en_us.json
BrailleBennett: The items will be automatically removed from other language files by Weblate, since en_us is the source language.

boholder (approved):

PR #399, opened 2024-12-01
Add an object tracker (POI target switching)

### LINKED ISSUE #186: POI Locking: Make it easier for players to lock what they want to lock.
Now with only the closest entity can be locked on, some players have reported that they can't lock on the Ender Dragon or the Shulker.

Some ideas:

* Two additional keys, one for locking friendly creatures and players, and another for locking onto hostile mobs. (We’ll keep the original locking key to mark on blocks.) Three locking keys for three types of target.

* One additional key for switching loop between three locking types.

- [ ] Add a key combination for toggle automatically refresh locking, to lock the nearest entity. It helps in battle with monsters.
- [x] #208

### LINKED ISSUE #341: Add target switching for POI and lockon
Original suggestion by @techexplorers on [Discord](https://discord.com/channels/843921862194364468/1250799911491080285/1282348901524967629)

We already have all the data needed for this feature so it shouldn't be that hard to add.

# Idea

Add keybinds (like CTRL + PGUP and CTRL + PGDOWN) for switching between target groups like friendly mobs, hostile mobs, functional blocks etc (also include groups from #267 when it gets implemented).
After selecting a group, add keybinds (like PGUP and PGDOWN) to go through targets in that group (both group and target selection would be narrated obviously). When selecting a target, maybe play a one time sound to indicate where is the target.
Finally, add a keybind (like HOME) to continuously play a repetitive sound that gets quicker the closer you get to the target. Additionally, when using lockon, instead of locking onto the nearest target, lock onto the currently selected target.

Molitvan:
Closes #341 

This feature will do everything outlined in #341 allowing players to get a much better idea of their surroundings and give players control over what the locking feature targets.

# Changelog

## New Features
- Added an Object Tracker. Find the changelog section on it below or learn more [on the docs](https://docs.mcaccess.org/features/#object-tracker)

## Object tracker

Gives the player a better idea of what is around them and allows for controlling what other POI features like locking target.
Keybinds:
- `PGUP`: previous object
- `PGDOWN`: next object
- `CTRL` + `PGUP`: previous group
- `CTRL` + `PGDOWN`: next group
- `HOME`: narrate current object
- `END`: target nearest object

## Feature Updates
- Locking feature's speak relative distance config is deprecated. It has been replaced by a new setting that affects both the locking feature and the object tracker.


common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
```diff
@@ -96,41 +101,83 @@ public class POIBlocks {
...
+            blockAlreadyInGroup = world.getBlockState(pos).getBlock() == world.getBlockState(posInGroup).getBlock();
+            if (blockAlreadyInGroup) break;
+        }
+
+        return !world.getBlockState(pos).isAir() && !blockAlreadyInGroup;
+    }
+
     @SuppressWarnings("unchecked")
-    private final POIGroup<BlockPos>[] groups = new POIGroup[] {
+    public final POIGroup<BlockPos>[] groups = new POIGroup[] {
```
boholder: I bet extracting groups into separate class fields isn't what TSG wants.
For me, initializing this `POIGroup<BlockPos>[] groups` within static block is better. For example:
```java
static {
        POIGroup<BlockPos> oreGroup = ...
        POIBlocks.groups = {
            markedGroup,
            oreGroup,
            ......
        }
    }
```
But I noticed that you change the `groups` back to non-static, so can't do this for now.
Special groups like `markedGroup` does deserve an independent class field, while the others don't.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
```diff
@@ -28,7 +33,7 @@
 @Slf4j
 public class POIBlocks {
     @Getter
-    private static final POIBlocks INSTANCE = new POIBlocks();
+    private static final POIBlocks instance = new POIBlocks();
     private ClientPlayerEntity player;
```
boholder: IntelliJ says this can be converted into local var

common/src/main/java/org/mcaccess/minecraftaccess/utils/KeyBindingsHandler.java
```diff
@@ -57,6 +58,11 @@ public class KeyBindingsHandler {
     public KeyBinding mouseSimulationScrollUpKey;
     public KeyBinding mouseSimulationScrollDownKey;
 
+    public KeyBinding objectTrackerNextItem;
+public KeyBinding objectTrackerPreviousItem;
```
boholder: fix the indentation

common/src/main/java/org/mcaccess/minecraftaccess/config/config_maps/POIConfigMap.java
```diff
@@ -27,6 +35,8 @@ public static POIConfigMap buildDefault() {
         m.poiEntitiesConfigMap = POIEntitiesConfigMap.buildDefault();
         m.poiLockingConfigMap = POILockingConfigMap.buildDefault();
         m.poiMarkingConfigMap = POIMarkingConfigMap.buildDefault();
+        
+        m.setSpeakTargetPosition(false);
```
boholder: use `m.speakTargetPosition = false`, and remove the previous blank line

common/src/main/java/org/mcaccess/minecraftaccess/config/config_maps/POIConfigMap.java
```diff
@@ -18,6 +23,9 @@ public class POIConfigMap {
     @SerializedName("Entities/Blocks Marking")
     private POIMarkingConfigMap poiMarkingConfigMap;
 
+    @SerializedName("Speak relative position of targets")
+    private boolean speakTargetPosition;
```
boholder: init the value with  `= false;`

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
```diff
@@ -212,7 +263,8 @@ private void checkBlock(BlockPos blockPos, int val) {
         }
 
         for (POIGroup<BlockPos> group : groups) {
-            if (group.add(blockPos)) {
+            if (group.add(blockPos) && group != otherBlocksGroup) {
```
boholder: I don't understand why making other blocks group as exception

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
```diff
@@ -96,41 +101,83 @@ public class POIBlocks {
...
             ),
+            otherBlocksGroup, // This group should always be at the end of this list
     };
 
     private POIBlocks() {
         loadConfigurations();
     }
 
+    @Getter
+    private List<BlockPos> lastScanResults = new ArrayList<>();
```
boholder, line 70: These two class fields can be moved to the front of methods

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIGroup.java
```diff
@@ -1,29 +1,44 @@
...
+    private Supplier<String> nameSupplier;
     private final SoundEvent sound;
     private final float soundPitch;
 
     private final Predicate<T> predicate;
     private final List<T> items = new ArrayList<>();
 
-    public POIGroup(SoundEvent sound, float soundPitch, Predicate<T> predicate) {
+    public POIGroup(Supplier<String> nameSupplier, SoundEvent sound, float soundPitch, Predicate<T> predicate) {
+        this.nameSupplier = nameSupplier;
```
boholder: What's the benefit of getting the name dynamicly from a supplier instead of a static string?

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/ObjectTracker.java
```diff
@@ -0,0 +1,207 @@
...
+        speakDistance = map.isSpeakTargetPosition();
+    }
+
+    private void narrateCurrentObject(boolean interupt) {
+        if (checkAndSpeakIfAllGroupsEmpty()) return;
+
+        POIGroup<?> currentGroup = groups.get(currentGroupIndex);
+
+        if (currentGroup.isEmpty()) {
+            MainClass.speakWithNarrator("No objects in current group", interupt);
```
boholder: Should use an I18N text instead

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
```diff
@@ -208,16 +211,16 @@ private void unlock(boolean speak) {
...
     private void relock() {
-        for (POIGroup<Entity> group : POIEntities.getInstance().groups) {
-            Optional<Entity> nearest = group.getItems().stream()
-                    .min(Comparator.comparingDouble(entity -> WorldUtils.getClientPlayer().distanceTo(entity)));
-            if (nearest.map(this::lockOnEntity).orElse(false)) {
-                return;
-            }
+        Object target = ObjectTracker.getInstance().getCurrentObject();
+
+        if (target instanceof Entity) {
```
boholder:
```java
switch (target) {
            case Entity entity -> lockOnEntity(entity);
            case BlockPos blockPos -> lockOnBlock(blockPos);
            default -> {
                return;
            }
        }
```

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIEntities.java
```diff
@@ -107,12 +122,14 @@ public void update(boolean isMarking, Entity markedEntity) {
         if (minecraftClient.world == null) return;
         if (minecraftClient.currentScreen != null) return; //Prevent running if any screen is opened
 
-        log.debug("POIEntities started.");
-
         for (POIGroup<Entity> group : groups) {
             group.clear();
         }
 
+            List<Entity> currentScanResults = new ArrayList<>();
```
boholder: Fix the indentation

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
```diff
@@ -160,6 +209,8 @@ public void update(boolean isMarking, Block markedBlock) {
         checkBlock(pos, this.range);
         checkBlock(pos.up(), this.range);
 
+        lastScanResults = currentScanResults;
```
boholder, line 113: You sure assigning this after `currentScanResults = new ArrayList<>();`? Both two are empty lists now. Doesn't sounds like what the `lastScanResults` name means, I've thought it's for saving the previous results.

common/src/main/java/org/mcaccess/minecraftaccess/utils/KeyBindingsHandler.java
```diff
@@ -440,6 +446,34 @@ private void initializeOtherKeybindings() {
                 GLFW.GLFW_KEY_H,
                 OTHER_GROUP_TRANSLATION_KEY
         );
+
+        objectTrackerNextItem = new KeyBinding(
+                I18n.translate("minecraft_access.keys.object_tracker.next_item"),
+                InputUtil.Type.KEYSYM,
+                GLFW.GLFW_KEY_PAGE_DOWN,
+                I18n.translate("minecraft_access.point_of_interest.object_tracker.group_name")
```
boholder: These four translation keys, "minecraft_access.point_of_interest.object_tracker.group_name", don't exist

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/ObjectTracker.java
```diff
@@ -0,0 +1,207 @@
...
+        if (!entities.isEmpty() && !blocks.isEmpty()) {
+            double distanceToEntity = MinecraftClient.getInstance().player.distanceTo(entities.getFirst());
+            double distanceToBlock = MinecraftClient.getInstance().player.getEyePos().distanceTo(blocks.getFirst().toCenterPos());
+
+            if (distanceToEntity <= distanceToBlock) currentObject = entities.getFirst();
+            if (distanceToBlock < distanceToEntity) currentObject = blocks.getFirst();
+        }
+
+        if (!entities.isEmpty() || !blocks.isEmpty()) {
+            MainClass.speakWithNarrator("Targeting nearest object", true);
```
boholder: Should use an I18N text instead

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/ObjectTracker.java
```diff
@@ -0,0 +1,207 @@
...
+            MainClass.speakWithNarrator(message, interupt);
+            WorldUtils.playSoundAtPosition(SoundEvents.BLOCK_NOTE_BLOCK_BELL, 1, 1f, block.toCenterPos());
+        }
+    }
+
+    private void moveGroup(int step) {
+        if (checkAndSpeakIfAllGroupsEmpty()) return;
+
+        if ((currentGroupIndex + step) > (groups.size() - 1)) {
+            MainClass.speakWithNarrator(I18n.translate("minecraft_access.other.end_of_list"), true);
```
boholder: In fact you can make the group switching an infinity loop, with `index = (index + groups.size()) % groups.size()`. So you can save two translations "start of list" "end of list" and making the user experience a little better. Speaking of which, we can also apply this on slot groups in Inventory Control.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/ObjectTracker.java
```diff
@@ -0,0 +1,207 @@
...
+
+@Slf4j
+public class ObjectTracker {
+    @Getter
+    private static ObjectTracker instance = new ObjectTracker();
+
+    private Keystroke nextItemKeyPressed = new Keystroke(() -> KeyUtils.isAnyPressed(KeyBindingsHandler.getInstance().objectTrackerNextItem), Keystroke.TriggeredAt.PRESSED);
+    private Keystroke previousItemKeyPressed = new Keystroke(() -> KeyUtils.isAnyPressed(KeyBindingsHandler.getInstance().objectTrackerPreviousItem), Keystroke.TriggeredAt.PRESSED);
+    private Keystroke narrateCurrentObjectKeyPressed = new Keystroke(() -> KeyUtils.isAnyPressed(KeyBindingsHandler.getInstance().objectTrackerNarrateCurrentObject), Keystroke.TriggeredAt.PRESSED);
+    private Keystroke targetNearestObjectKeyPressed = new Keystroke(() -> KeyUtils.isAnyPressed(KeyBindingsHandler.getInstance().targetNearestObject), Keystroke.TriggeredAt.PRESSED);
```
boholder: These four fields can be `final`

boholder (changes requested):
In running, the new feature does not take effect for either block or entity, and the POI Block disappears.The POI Entity is valid. Can you provide a simple sequence of key pressing for testing?

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
sgmedev: Extracting groups into separate fields is perfectly fine, and IMO is the best way to handle this. I actually generally don’t like static blocks because often they end up just duplicating code by separating the declaration and assignment where these could be combined into one statement.

Now that I think about it, an enum might be even better for this if most groups need to be accessed separately since enums let you both iterate over all their variants and get individual variants without any duplication or spreading stuff out at all, you could define such an enum something like this:
```java
enum BlockGroups {
    FLUIDS(
        SoundEvents.BLOCK_NOTE_BLOCK_BIT.value(),
        2f,
        pos -> this.detectFluidBlocks && world.getBlockState(pos).getBlock() instanceof FluidBlock && PlayerUtils.isNotInFluid() && world.getFluidState(pos).getLevel() == 8
    ),
    FUNCTIONAL(
        SoundEvents.BLOCK_NOTE_BLOCK_BIT.value(),
        2f,
        pos -> world.getBlockState(pos).getBlock() instanceof ButtonBlock || world.getBlockState(pos).getBlock() instanceof LeverBlock || poiBlockPredicates.stream().anyMatch(p -> p.test(world.getBlockState(pos)))
    );

    public final POIGroup<BlockPos> group;

    BlockGroups(SoundEvent sound, float soundPitch, Predicate<BlockPos> predicate) {
        group = new POIGroup<>(sound, soundPitch, predicate);
    }
}
```

boholder:
I suggest you making this PR to be merged before making additional POI changes, for avoiding creating a big PR.

Molitvan:
> In running, the new feature does not take effect for either block or entity, and the POI Block disappears.The POI Entity is valid. Can you provide a simple sequence of key pressing for testing?
Weird. Works mostly fine for me and the testers. Make sure to increase your POI detecting range because the objects will disappear from the list if they're out of range.

Example test:
- Summon a pig and a sheep, place a crafting table and a furnace
- Use CTRL + PGUP/PGDOWN to scroll through groups. In this case, the groups should be passive entities, blocks with interface and other blocks.
- Use PGUP and PGDOWN to scroll to objects in a group. While scrolling, you should hear the object's name and a positional sound comming from where the object is.
- Pressing HOME should repeat the narration and sound
- Press END to target the nearest object
- At any point, use the locking feature to lock onto the current object

Molitvan:
I did everything except make the group list wrap.

> IntelliJ says this can be converted into local var

I didn't change that line. Don't know why Git thinks I did.

> You sure assigning this after currentScanResults = new ArrayList<>();? Both two are empty lists now. Doesn't sounds like what the lastScanResults name means, I've thought it's for saving the previous results.

currentScanResults gets populated in the checkBlock methods. The current scan is the one currently running in the checkBlock methods, and last scan is the last completed scan.

> In fact you can make the group switching an infinity loop, with index = (index + groups.size()) % groups.size(). So you can save two translations "start of list" "end of list" and making the user experience a little better. Speaking of which, we can also apply this on slot groups in Inventory Control.

I usually don't really like wrapping lists, but in this case it may actually be a good idea. Also please don't apply this to inventory controls because it will make it very confusing. What we should probably do for the inventory controls though is add keys for snapping to the most left/right and most up/down slots in the current slot group.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
Molitvan: > Extracting groups into separate fields is perfectly fine, and IMO is the best way to handle this. I actually generally don’t like static blocks because often they end up just duplicating code by separating the declaration and assignment where these could be combined into one statement.
> 
> Now that I think about it, an enum might be even better for this if most groups need to be accessed separately since enums let you both iterate over all their variants and get individual variants without any duplication or spreading stuff out at all, you could define such an enum something like this:
> 
> ```java
> enum BlockGroups {
>     FLUIDS(
>         SoundEvents.BLOCK_NOTE_BLOCK_BIT.value(),
>         2f,
>         pos -> this.detectFluidBlocks && world.getBlockState(pos).getBlock() instanceof FluidBlock && PlayerUtils.isNotInFluid() && world.getFluidState(pos).getLevel() == 8
>     ),
>     FUNCTIONAL(
>         SoundEvents.BLOCK_NOTE_BLOCK_BIT.value(),
>         2f,
>         pos -> world.getBlockState(pos).getBlock() instanceof ButtonBlock || world.getBlockState(pos).getBlock() instanceof LeverBlock || poiBlockPredicates.stream().anyMatch(p -> p.test(world.getBlockState(pos)))
>     );
> 
>     public final POIGroup<BlockPos> group;
> 
>     BlockGroups(SoundEvent sound, float soundPitch, Predicate<BlockPos> predicate) {
>         group = new POIGroup<>(sound, soundPitch, predicate);
>     }
> }
> ```

I have no idea how that code works lol

Edit: I know how to use enums, but they're usually not written like this, at least not in other languages

Molitvan:
> I suggest you making this PR to be merged before making additional POI changes, for avoiding creating a big PR.
Agreed

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
sgmedev: Java enums are able to have constructors, fields, and methods like any other class, all you have to do is put a semicolon after the last variant. When you define a constructor, you have to put the constructor parameters in parentheses after every variant as seen in my example. Every enum variant is an instance of the enum, so in this example if you wanted to access the fluids group, for example, you would write `BlockGroups.FLUIDS.group`. You can also use `BlockGroups.values()` to get an array with all the values which can be iterated over.

boholder:
> > IntelliJ says this can be converted into local var
> 
> I didn't change that line. Don't know why Git thinks I did.

Help all of us, fix it with `alt` + `enter`, IntelliJ will do it automatically, thanks.

> > You sure assigning this after currentScanResults = new ArrayList<>();? Both two are empty lists now. Doesn't sounds like what the lastScanResults name means, I've thought it's for saving the previous results.
> 
> currentScanResults gets populated in the checkBlock methods. The current scan is the one currently running in the checkBlock methods, and last scan is the last completed scan.

As long as you confirm the logic is right, the naming of these variables is good to me.

boholder:
@Molitvan There are unresolved questions, would we discuss about them before reviewing?

boholder:
@Molitvan There are some places that haven't been settled:

1. [Change poi groups to enum](https://github.com/minecraft-access/minecraft-access/pull/399#discussion_r1920828276), maybe @TheSuperGamer20578 can directly provide the code for you to patch.
2. https://github.com/minecraft-access/minecraft-access/pull/399#discussion_r1912351998
3. Maybe duplicate with 1: https://github.com/minecraft-access/minecraft-access/pull/399#discussion_r1912352280
4. https://github.com/minecraft-access/minecraft-access/pull/399#discussion_r1912354565

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
boholder: > I don't understand why making other blocks group as exception

: Because if it puts everything in like the other groups, you'll end up with a billion different things in there and it'll turn into an unusable mess

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIGroup.java
boholder: > What's the benefit of getting the name dynamicly from a supplier instead of a static string?

: I couldn't get i18n to work properly when using a static string. Also I didn't test this, but I think changing the language also wouldn't work when using a static string

Molitvan:
Is this done now or are there still things that need changes? I saw discussion about switching to enums on Discord.

boholder:
> Is this done now or are there still things that need changes?

Still WIP in my perspective, see https://github.com/minecraft-access/minecraft-access/compare/dev...boholder:minecraft-access:poi-object-tracker

Molitvan:
> > Is this done now or are there still things that need changes?
> 
> Still WIP in my perspective, see [dev...boholder:minecraft-access:poi-object-tracker](https://github.com/minecraft-access/minecraft-access/compare/dev...boholder:minecraft-access:poi-object-tracker)

Ok from what I can tell it's like I said on Discord. It's feature-complete, but there's a bunch of refactoring left to do.

Oh and btw if I can help with anything, just say.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
```diff
@@ -96,46 +96,85 @@ public class POIBlocks {
...
             -5f,
             pos -> oreBlockPredicates.stream().anyMatch(p -> p.test(world.getBlockState(pos)))
     );
 
+    private final POIGroup<BlockPos> otherBlocksGroup = new POIGroup<>(
+            () -> I18n.get("minecraft_access.point_of_interest.group.otherBlocks"),
+        pos -> otherBlocksFilter(pos)
+    );
+
+    private boolean otherBlocksFilter(BlockPos pos) {
```
boholder: @Molitvan Can you explain the idea of having an "other blocks group"?
I can't get this filter method's purpose: 1. it involves its "parent" - "otherBlocksGroup" 2. it compares block type instead of block position

Molitvan: > @Molitvan Can you explain the idea of having an "other blocks group"? I can't get this filter method's purpose: 1. it involves its "parent" - "otherBlocksGroup" 2. it compares block type instead of block position

The other blocks group contains the closest of every block that wasn't picked up by any other POI group around the player. This is useful when trying to find something that is not considered a POI, for example until we make a proper trees category, this is a decent way to find trees. It compares block type because again, it contains only one of every block type around the player.

boholder: > The other blocks group contains the closest of every block that wasn't picked up by any other POI group around the player.

Thanks, I added the description to code

```java
    /**
     * The other blocks group contains the closest of every block
     * that wasn't picked up by any other POI group around the player.
     * This is useful when trying to find something that is not considered a POI,
     * for example until we make a proper trees category, this is a decent way to find trees.
     */
    private final POIGroup<BlockPos> otherBlocksGroup = new POIGroup<>(
        "minecraft_access.point_of_interest.group.otherBlocks",
            this::blockIsNotAirAndNotContainedInGroupYet
    );

    private boolean blockIsNotAirAndNotContainedInGroupYet(BlockPos pos) {
        BlockState state = WorldUtils.getBlockState(pos);
        Set<Block> groupContainedBlockTypes = otherBlocksGroup.getItems().stream()
                .map(p -> WorldUtils.getBlockState(p).getBlock()).collect(Collectors.toSet());
        boolean blockAlreadyInGroup = groupContainedBlockTypes.contains(state.getBlock());
        return !state.isAir() && !blockAlreadyInGroup;
    }
```

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
boholder: @TheSuperGamer20578 `POIGroup` can't be enum since I want to dynamically create its instances from user custom config in the future. I managed to collect builtin groups into enum

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
boholder: @Molitvan Wait, it seems that the content of other blocks group isn't included in "lastScanResults" for locking feature, so can I make conclusion that the "otherBlocksGroup" is unused, for now, in scope of this PR? Maybe deleting it is better?

Molitvan: > @Molitvan Wait, it seems that the content of other blocks group isn't included in "lastScanResults" for locking feature, so can I make conclusion that the "otherBlocksGroup" is unused, for now, in scope of this PR? Maybe deleting it is better?

The other blocks group is fully complete and being used. This gives me the impression that you didn't even run the PR build. Could you please do that? You can't properly test features by just looking at the code.

boholder: > The other blocks group is fully complete and being used.

I'm checking the code in IDE. The "otherBlocksGroup" has three usages and all of three are for modifying group content, the only "get" usage is in its filter method.

Molitvan: > > The other blocks group is fully complete and being used.
> 
> I'm checking the code in IDE. The "otherBlocksGroup" has three usages and all of three are for modifying group content, the only "get" usage is in its filter method.

That's right, but it is included in the POIBlocks groups list, which means it's included in the object tracker. The get method isn't used directly, but it is used when selected in the object tracker, just like every other group.

Again, I request you please run the build, because it's impossible to properly review features without doing that.

Open any world and stand on the ground. You'll notice the other blocks group at the bottom of the object tracker, fully functioning. That looks very used to me.

boholder: > That's right, but it is included in the POIBlocks groups list, which means it's included in the object tracker. The get method isn't used directly, but it is used when selected in the object tracker, just like every other group.

You're right, sorry for not fully reading the code.

boholder (approved):
I tested all the POI features and new tracker features in dev env.

PR #400, opened 2024-12-08
Fixed SAPI not triggering when no supported screen reader

BrailleBennett:
[//]: # (The changelog will be added into CHANGELOG.md file automatically by fast-forward workflow.)

[//]: # (Please delete unrelevant changelog sections.)

### New Features
- When you do not have NVDA running on windows, Tolk will now correctly fall back to the built-in SAPI support instead of not narrating anything.
- You can now use Braille only on windows if desired for mod narration output via setting your screen reader


BrailleBennett:
Everything works besides building the fabric jar directly :(

BrailleBennett:
An update for those reading this in future. I got it to a state where it works in the dev environment. The final roadblock is that I somehow have to take the downloaded jar file and unzip/unpack it inside of the mod in order to access the tolk wrapper. I'm not sure how to best go about doing it, and it has been pretty low priority overall. As of now, there will definitely be some merge conflicts because of the large change to how gradle runs in this project and the inclusion of the gradle versions catalog. I will try and return to this PR in future, but if you have any ideas, please feel free to continue this PR yourself.

BrailleBennett:
No more focus should be placed on this PR in favor of focusing on #568

PR #404, opened 2024-12-11
renamed packages to match org structure

### LINKED ISSUE #401: Update package name
The current package name `com.github.khanshoaib3.minecraft_access` is no longer valid since we have moved to an org, it should be renamed to `io.github.minecraft_access.minecraft_access`. This should not be a breaking change since we do not currently expose any Java API.

BrailleBennett:
This changes the file structure and package imports project-wide. I tested to make sure that Fabric and NeoForge ran and built, but someone may want to make sure there aren't any paths I messed up in the build files that would no longer match because of the restructure.
Closes #401


Molitvan:
This will break every PR that adds new classes, so both of my PRs. I suggest merging everything else first, then updating this to include the new classes, then merging it

BrailleBennett:
> This will break every PR that adds new classes, so both of my PRs. I suggest merging everything else first, then updating this to include the new classes, then merging it

One of your PRs is still a draft though. That would put the responsibility on you to fix it since mine was ready to be merged first. I can wait and merge until CI is done and everything else done is merged, but unless your PR is done by then I personally don't feel like I should have to wait for it to be, as there's no direct relationship between the two, just a miled inconvenience for you to fix it.

sgmedev (changes requested):
Please consider https://github.com/minecraft-access/minecraft-access/issues/401#issuecomment-2549855516

PR #410, opened 2024-12-22
Rename repo package

### LINKED ISSUE #401: Update package name

BrailleBennett:
.


gradle.properties
```diff
@@ -11,12 +11,12 @@ enabled_platforms=fabric,neoforge
...
+maven_group=org.mcaccess.minecraftaccess
 
 fabric_loader_version=0.15.11
 fabric_api_version=0.110.0+1.21.1
 
 architectury_api_version = 13.0.8
 
 neoforge_version=21.1.4
-yarn_mappings_patch_neoforge_version=1.21+build.4
+yarn_mappings_patch_neoforge_version=1.21+build.4
```
sgmedev: No newline at end of file

sgmedev (changes requested):

sgmedev:
Fabric and NeoForge also still need to be renamed

sgmedev (approved):

PR #414, opened 2024-12-24
Prerelease-fixes/v1.9.0-beta.1

sgmedev:
Fix issues found in v1.9.0-beta.1

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/features/PlayerWarnings.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/features/PlayerWarnings.java
@@ -19,7 +19,7 @@ public class PlayerWarnings {
     private boolean isHealthBelowSecondThreshold;
     private boolean isFoodBelowThreshold;
     private boolean isAirBelowThreshold;
-    private  boolean isFrostAboveThreshold;
+    private boolean isFrostAboveThreshold;
 
     private boolean playSound;
     private double firstHealthThreshold;
@@ -39,25 +39,23 @@ public class PlayerWarnings {
     }
 
     public void update() {
-        try {
-            minecraftClient = MinecraftClient.getInstance();
-            if (minecraftClient == null) return;
-            if (minecraftClient.player == null) return;
-            if (minecraftClient.currentScreen != null) return;
+        minecraftClient = MinecraftClient.getInstance();
+        if (minecraftClient == null) return;
+        if (minecraftClient.player == null) return;
+        if (minecraftClient.currentScreen != null) return;
 
-            loadConfigurations();
+        loadConfigurations();
 
-            double maxHealth = Math.round((minecraftClient.player.getMaxHealth() / 2.0) * 10.0) / 10.0;
-            double maxHunger = Math.round((20 / 2.0) * 10.0) / 10.0;
-            double maxAir = Math.round((minecraftClient.player.getMaxAir() / 20.0) * 10.0) / 10.0;
-            double frostExposurePercent = Math.round((minecraftClient.player.getFreezingScale() * 100.0) * 10.0) / 10.0;
+        double maxHealth = Math.round((minecraftClient.player.getMaxHealth() / 2.0) * 10.0) / 10.0;
+        double maxHunger = Math.round((20 / 2.0) * 10.0) / 10.0;
+        double maxAir = Math.round((minecraftClient.player.getMaxAir() / 20.0) * 10.0) / 10.0;
+        double frostExposurePercent = Math.round((minecraftClient.player.getFreezingScale() * 100.0) * 10.0) / 10.0;
 
-            healthWarning(PlayerUtils.getHearts(), maxHealth);
+        healthWarning(PlayerUtils.getHearts(), maxHealth);
+        if (!minecraftClient.player.isCreative()) {
             hungerWarning(PlayerUtils.getHunger(), maxHunger);
             airWarning(Math.round((minecraftClient.player.getAir() / 20.0) * 10.0) / 10.0, maxAir);
             frostWarning(frostExposurePercent);
-        } catch (Exception e) {
-            log.error("An error occurred in PlayerWarnings.", e);
         }
     }
 
@@ -116,7 +114,7 @@ public class PlayerWarnings {
     }
 
     private void frostWarning(double frostExposurePercent) {
-        if (minecraftClient.player == null || minecraftClient.player.isCreative()) return;
+        if (minecraftClient.player == null) return;
 
         if (frostExposurePercent >= frostThreshold && frostExposurePercent < 100 && !isFrostAboveThreshold) {
             isFrostAboveThreshold = true;
--- a/common/src/main/java/org/mcaccess/minecraftaccess/features/PlayerWarnings.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/features/PlayerWarnings.java
@@ -116,7 +116,7 @@ public class PlayerWarnings {
     }
 
     private void frostWarning(double frostExposurePercent) {
-        if (minecraftClient.player == null) return;
+        if (minecraftClient.player == null || minecraftClient.player.isCreative()) return;
 
         if (frostExposurePercent >= frostThreshold && frostExposurePercent < 100 && !isFrostAboveThreshold) {
             isFrostAboveThreshold = true;
--- a/common/src/main/java/org/mcaccess/minecraftaccess/features/PlayerWarnings.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/features/PlayerWarnings.java
@@ -54,7 +54,7 @@ public class PlayerWarnings {
 
             healthWarning(PlayerUtils.getHearts(), maxHealth);
             hungerWarning(PlayerUtils.getHunger(), maxHunger);
-            airWarning(PlayerUtils.getAir(), maxAir);
+            airWarning(Math.round((minecraftClient.player.getAir() / 20.0) * 10.0) / 10.0, maxAir);
             frostWarning(frostExposurePercent);
         } catch (Exception e) {
             log.error("An error occurred in PlayerWarnings.", e);
--- a/common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
@@ -1,5 +1,6 @@
 package org.mcaccess.minecraftaccess.features.point_of_interest;
 
+import net.minecraft.entity.ItemEntity;
 import org.mcaccess.minecraftaccess.MainClass;
 import org.mcaccess.minecraftaccess.config.config_maps.POILockingConfigMap;
 import org.mcaccess.minecraftaccess.utils.KeyBindingsHandler;
@@ -271,11 +272,16 @@ public class LockingHandler {
         unlock(false);
         lockedOnEntity = entity;
 
-        String narration = NarrationUtils.narrateEntity(entity);
-        if (this.speakDistance) {
-            narration += " " + NarrationUtils.narrateRelativePositionOfPlayerAnd(entity.getBlockPos());
+        String toSpeak = "";
+        if(entity instanceof ItemEntity) {
+         toSpeak+=I18n.translate("minecraft_access.point_of_interest.locking.dropped_item", NarrationUtils.narrateEntity(entity));
+        } else {
+            toSpeak += NarrationUtils.narrateEntity(entity);
         }
-        MainClass.speakWithNarrator(I18n.translate("minecraft_access.point_of_interest.locking.locked", narration), true);
+        if (this.speakDistance) {
+            toSpeak += " " + NarrationUtils.narrateRelativePositionOfPlayerAnd(entity.getBlockPos());
+        }
+        MainClass.speakWithNarrator(I18n.translate("minecraft_access.point_of_interest.locking.locked", toSpeak), true);
         return true;
     }
 
--- a/common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
@@ -306,7 +306,7 @@ public class LockingHandler {
             default -> position.toCenterPos();
         };
 
-        lockedOnBlock = new BlockPos3d(absolutePosition);
+        lockedOnBlock = new BlockPos3d(position, absolutePosition);
 
         String blockDescription = NarrationUtils.narrateBlock(lockedOnBlock, "");
         if (this.speakDistance) {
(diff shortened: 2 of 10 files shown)


common/src/main/java/org/mcaccess/minecraftaccess/features/PlayerWarnings.java
boholder, line 1: I still prefer moving these logic into PlayerUtils... I mean, why not? We can't expect when and where they will be reused.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
```diff
@@ -23,10 +25,8 @@
 import net.minecraft.util.math.Vec3d;
 import net.minecraft.item.BowItem;
 
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
boholder, line 28: Since we'll have linting, let me explain more about this. I agree with the top answer of SOF question:
https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports
https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html

IntelliJ will wrap the import part BTW. One potential problem is, a long import list will makes blind developers take time to scroll over if they are reading the code with non-IDE tools, like, GitHub website.

common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
```diff
@@ -111,6 +113,9 @@ public static String narrateEntity(Entity entity) {
             text = I18n.translate("minecraft_access.other.entity_with_equipments", values);
         }
 
+        if (entity instanceof ItemEntity itemEntity && itemEntity.isOnGround() || entity instanceof PersistentProjectileEntity projectile && projectile.pickupType.equals(PersistentProjectileEntity.PickupPermission.ALLOWED))
```
boholder: Seems like the cast variable "itemEntity" isn't used inside the `if` scope, can we remove it?

boholder: How about assigning one or three (`isDroppedItem`, `isPickableProjectile`, `isPickable`) boolean variable for this long condition?

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
```diff
@@ -207,12 +208,11 @@ private void unlock(boolean speak) {
...
-            TreeMap<Double, Entity> map = group.getEntities();
-            if (!map.isEmpty()) {
-                Entity entity = map.firstEntry().getValue();
-                if (lockOnEntity(entity)) return;
+        for (POIGroup<Entity> group : POIEntities.getInstance().groups) {
+            Optional<Entity> nearest = group.getItems().stream()
+                    .min(Comparator.comparingDouble(entity -> WorldUtils.getClientPlayer().distanceTo(entity)));
+            if (nearest.map(this::lockOnEntity).orElse(false)) {
+                return;
             }
```
boholder, line 216: Yeah, much clearer, I haven't realized that the distance info should be delayed to be produced just before where they're used.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
```diff
@@ -286,47 +286,25 @@ public boolean lockOnEntity(Entity entity) {
...
+                .ifPresent(this::lockOnBlock);
     }
 
-    private void lockOnBlock(Vec3d position) {
+    private void lockOnBlock(BlockPos position) {
         unlock(false);
 
-        BlockState blockState = WorldUtils.getClientWorld().getBlockState(WorldUtils.blockPosOf(position));
+        BlockState blockState = WorldUtils.getClientWorld().getBlockState(position);
         entriesOfLockedOnBlock = blockState.getEntries();
```
boholder, line 294: So... the [WorldUtils.blockPosOf()](https://github.com/minecraft-access/minecraft-access/blob/91d139b94900e5dc82a8c9a7e50e4572787830d5/common/src/main/java/org/mcaccess/minecraftaccess/utils/WorldUtils.java#L20) is a fix for breaking changes of original code. https://github.com/minecraft-access/minecraft-access/commit/ed6f890d27484304cb4aad7a117216a1ede9a384
Make sure the locking feature on things like door or button is accurate.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
```diff
@@ -88,63 +92,91 @@ public class POIBlocks {
...
+                    SoundEvents.BLOCK_NOTE_BLOCK_BIT.value(),
+                    2f,
+                    pos -> world.getBlockState(pos).getBlock() instanceof ButtonBlock || world.getBlockState(pos).getBlock() instanceof LeverBlock || poiBlockPredicates.stream().anyMatch(p -> p.test(world.getBlockState(pos)))
+            ),
+            new POIGroup<BlockPos>(// Blocks with interface
+                    SoundEvents.BLOCK_NOTE_BLOCK_BANJO.value(),
+                    0f,
+                    pos -> world.getBlockState(pos).createScreenHandlerFactory(world, pos) != null
+            ),
+    };
```
boholder, line 128: How about extracting these `POIGroup` instances into their own variable (inside the static block)? To make the meaning of each group more clear.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
```diff
@@ -179,49 +211,22 @@ private void checkBlock(BlockPos blockPos, int val) {
...
-            results.add(builtInGroups.get("marked").getBlocks(true));
-            return results;
-        }
-
-        for (POIGroup group : builtInGroups.values()) {
-            results.add(group.getBlocks(true));
+    public @UnmodifiableView List<BlockPos> getLockingCandidates() {
+        if (isMarking && POIMarkingConfigMap.getInstance().isSuppressOtherWhenEnabled()) {
+            return markedGroup.getItems();
         }
```
boholder, line 228: Want a little `else` for clarifying the `if-condition`?

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIEntities.java
```diff
@@ -38,100 +36,104 @@ public class POIEntities {
...
+                    SoundEvents.BLOCK_NOTE_BLOCK_IRON_XYLOPHONE.value(),
+                    1f,
+                    entity -> entity instanceof VehicleEntity
+            ),
+            new POIGroup<Entity>(// Items
+                    SoundEvents.BLOCK_METAL_PRESSURE_PLATE_CLICK_ON,
+                    2f,
+                    entity -> entity instanceof ItemEntity itemEntity && itemEntity.isOnGround() || entity instanceof PersistentProjectileEntity projectile && projectile.pickupType.equals(PersistentProjectileEntity.PickupPermission.ALLOWED)
+            ),
+    };
```
boholder, line 86: Same as POI Block, consider extract the groups.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIEntities.java
```diff
@@ -38,100 +36,104 @@ public class POIEntities {
...
-            List<Entity> entities = minecraftClient.world.getOtherEntities(minecraftClient.player, scanBox);
+        for (POIGroup<Entity> group : groups) {
+            group.clear();
+        }
 
-            if (onPOIMarkingNow && POIMarkingConfigMap.getInstance().isSuppressOtherWhenEnabled()) {
-                POIGroup passiveGroup = builtInGroups.get("passive");
-                POIGroup hostileGroup = builtInGroups.get("hostile");
+        Box scanBox = minecraftClient.player.getBoundingBox().expand(range, range, range);
+        List<Entity> entities = minecraftClient.world.getOtherEntities(minecraftClient.player, scanBox);
```
boholder, line 117: #TIL

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIEntities.java
```diff
@@ -38,100 +36,104 @@ public class POIEntities {
...
 
-            for (POIGroup group : builtInGroups.values()) {
-                for (Entity e : entities) {
-                    if (group.checkAndAddEntity(e)) {
-                        this.playSoundAt(e.getBlockPos(), group.sound, group.soundPitch);
-                        entities.remove(e);
-                    }
+        for (POIGroup<Entity> group : groups) {
+            for (Entity entity : group.getItems()) {
+                if (isMarking && POIMarkingConfigMap.getInstance().isSuppressOtherWhenEnabled() && !(marked == null || marked.isInstance(entity))) {
```
boholder, line 125: Sorry but I can't understand the meaning of the `!(marked == null || marked.isInstance(entity))`, could you clarify it with a boolean variable?

common/src/main/java/org/mcaccess/minecraftaccess/utils/condition/IntervalKeystroke.java
```diff
@@ -46,4 +46,13 @@ public boolean canBeTriggered() {
     protected boolean otherTriggerConditions() {
         return true;
     }
+
+    @Override
+    public void updateStateForNextTick() {
```
boholder, line 51: Wait, I haven't saw the usage of it... can you provide the commit that contains this change? I want to know the purpose.
Wait, is that left fixing? In the access menu shortcuts bar PR?

boholder (changes requested):

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
sgmedev: Alot of stuff will need reformatting when the linter is setup. I don't think it's worth worrying about minor things like this until then.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
sgmedev: That would make it significantly harder to maintain because then some changes such as adding or removing a group will need to be made in multiple places. Order is also significant here so it would be less readable because you would have to reference the order in the array which is seperate at the same time. Is the comment there really that insufficient?

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIEntities.java
sgmedev: That would make it significantly harder to maintain because then some changes such as adding or removing a group will need to be made in multiple places. Order is also significant here so it would be less readable because you would have to reference the order in the array which is seperate at the same time. Is the comment there really that insufficient?

common/src/main/java/org/mcaccess/minecraftaccess/utils/condition/IntervalKeystroke.java
sgmedev: IntervalKeystroke had a problem where if you double tap a key within the interval it will just ignore it which is not good. This fixes it by making the interval ready on key down which means it will always execute no matter how quickly you tap the key. A quick Git blame or scrolling through the commit log would tell you that this change was made in commit 14a284bd926c255011f556bf05fd7c2f1f0a2fff.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
sgmedev: An else statement would be superfluous here and would typically be regarded as less clear because it makes the code more deeply nested.

common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
sgmedev: It's used in the condition.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIBlocks.java
boholder: Well, as long as the assignment is inside the static block, I can tell it's sort of initialization, not the runtime result.

common/src/main/java/org/mcaccess/minecraftaccess/utils/condition/IntervalKeystroke.java
boholder: Sorry... I'm on my phone, does the logic same as the fixing I made for the... previous problem?

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIEntities.java
sgmedev: That checks if the entity is not an instance of the marked entity and thus should be silenced. It also has a simple null check thrown in because marked is nullable.

common/src/main/java/org/mcaccess/minecraftaccess/utils/condition/IntervalKeystroke.java
boholder: Back to the PC now, IIRC it's: https://github.com/minecraft-access/minecraft-access/pull/349#issuecomment-2480888163

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIEntities.java
boholder: if `POI marking on` and `suppress enabled` and, `marked != null` and `the entity is NOT the marked type`, yeah, I understand now, does IntelliJ automatically suggested extracting the `not`? I remember I learned the knowledge in university... how about adding the `not` back?

common/src/main/java/org/mcaccess/minecraftaccess/features/PlayerWarnings.java
BrailleBennett: Like I said to you on the discord, this is out of scope for this pull request. You should have raised this issue earlier.

common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
BrailleBennett: Because it's literally only used once. Adding it to a variable is extra and doesn't really help the readability of the code.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
BrailleBennett: Yes, I tested partial blocks and they all worked.

common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
BrailleBennett: I added it.

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/POIEntities.java
sgmedev: Intellij did not have any suggestions when I wrote it. I'm also not quite sure what you mean.

boholder (approved):

BrailleBennett (approved):

PR #419, opened 2024-12-27
Updated dependencies and buildtools (gradle 8.12)

BrailleBennett:
- Gradle 8.11.1 > Gradle 8.12
- Fabric API 0.112.2 > 0.113.0
- NeoForge 21.4.33-beta > 21.4.44-beta

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ maven_group=org.mcaccess.minecraftaccess
 architectury_api_version=15.0.1
 fabric_yarn_version=1.21.4+build.2:v2
 fabric_loader_version=0.16.9
-fabric_api_version=0.112.2+1.21.4
-neoforge_version=21.4.33-beta
+fabric_api_version=0.113.0+1.21.4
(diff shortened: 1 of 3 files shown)


boholder (approved):
Here are thoughts behind these changes:
1. the command inside `gradlew`, why it's changed before? It seems like a syntax error. 

just run
./gradlew wrapper
After changing the version in the gradle properties file and it touches the files gradlew, gradle.bat, and gradle.jar all on its own. I never interacted with them directly. It is advised to run gradle wrapper to insure the wrappers are up to date with the latest fixes from the gradle team.

2. fabric and neoforge versions changed but archi api not, why?

This is because archiAPI has no new version, I checked. I'm sure they will eventually, but for now they claim compatibility with the latest fabric API and NeoForge versions, so it's not really our concern.

3. What’s the point of following gradle’s latest version?

there's error logging improvements as well as 74 closed issues since 8.11
It's not like not updating makes it more stable, gradle has no such concept as LTS versions, it's rolling release.

PR #422, opened 2024-12-29
fix: the new "Mac Mouse Fix" doesn't have default value which causes NPE

### LINKED ISSUE #429: Inventory controls doesn't speak when moving to a slot with same contents
When using the inventory controls keys to move to slots with the same contents e.g. moving from an empty slot to another empty slot, inventory controls doesn't say anything. When speak focused slot changes is disabled, it doesn't narrate the inventories at all. This is not an issue in the latest stable version so I assume it was introduced after it's release.

boholder:
place holder, this pr doesn't worth a changelog I think?

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/config/config_maps/MouseSimulationConfigMap.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/config/config_maps/MouseSimulationConfigMap.java
@@ -16,7 +16,7 @@ public class MouseSimulationConfigMap {
     @SerializedName("Scroll Delay (in milliseconds)")
     private int scrollDelayInMilliseconds;
     @SerializedName("Mac Mouse Fix")
-	private Boolean macMouseFix = false;
+    private Boolean macMouseFix = false;
 
     private MouseSimulationConfigMap() {
     }
@@ -25,7 +25,7 @@ public class MouseSimulationConfigMap {
         MouseSimulationConfigMap m = new MouseSimulationConfigMap();
         m.enabled = true;
         m.scrollDelayInMilliseconds = 150;
-		m.macMouseFix = false;
+        m.macMouseFix = false;
 
         setInstance(m);
         return m;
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/mcaccess/minecraftaccess/config/config_maps/MouseSimulationConfigMap.java
```diff
@@ -15,8 +15,8 @@ public class MouseSimulationConfigMap {
     private boolean enabled;
     @SerializedName("Scroll Delay (in milliseconds)")
     private int scrollDelayInMilliseconds;
-    @SerializedName("mac mouse fix")
-	private Boolean macMouseFix;
+    @SerializedName("Mac Mouse Fix")
+	private Boolean macMouseFix = false;
```
sgmedev: This line is indented with a tab, it should be indented with four spaces.
```suggestion
    private Boolean macMouseFix = false;
```

sgmedev (changes requested):
Is the default not set on line 28 (which is also erroneously indented with a tab)?
```java
m.macMouseFix = false;
```

boholder:
> Is the default not set on line 28 (which is also erroneously indented with a tab)?
> 
> ```java
> 
> m.macMouseFix = false;
> 
> ```

Weird, it's formatted normally in the IDE. Do we encounter sort of invisible characters (I remember there are related attacking)

BrailleBennett (dismissed):
Disregard

BrailleBennett:
Ugh, wrong PR

BrailleBennett (changes requested):
You must put something in the PR description. If it is fully empty the CI will error.

sgmedev:
> > Is the default not set on line 28 (which is also erroneously indented with a tab)?
> > 
> > ```java
> > 
> > m.macMouseFix = false;
> > 
> > ```
> 
> Weird, it's formatted normally in the IDE. Do we encounter sort of invisible characters (I remember there are related attacking)

It's a tab character, it looks the same as four spaces in your current IDE configuration but is actually one character which you will notice when you try to highlight it. On GitHub it looks like stuff is indented wrongly because it displays a tab as eight spaces for some reason, this inconsistency in tab width between viewers is why it's imparative not to use tabs anywhere.

boholder:
> It's a tab character

So that's a good rule for linter.... replace tab with spaces

sgmedev (approved):

BrailleBennett (approved):

PR #427, opened 2025-01-01
fix: can't speak holding item (hotbar) in v1.9.0

### LINKED ISSUE #426: Can't speak holding item (hotbar) in v1.9.0
While the fabric version of mod doesn't have this issue.

boholder:
# Changelog

## Bug Fixes

- Fixed hotbar items not being spoken on the NeoForge platform

### DIFF
--- a/neoforge/src/main/java/org/mcaccess/minecraftaccess/neoforge/mixin/InGameHudMixin.java
+++ b/neoforge/src/main/java/org/mcaccess/minecraftaccess/neoforge/mixin/InGameHudMixin.java
@@ -1,0 +1,35 @@
+package org.mcaccess.minecraftaccess.neoforge.mixin;
+
+import net.minecraft.client.gui.DrawContext;
+import net.minecraft.client.gui.hud.InGameHud;
+import net.minecraft.item.ItemStack;
+import org.mcaccess.minecraftaccess.features.SpeakHeldItem;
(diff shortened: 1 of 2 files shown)


boholder:
~~Use method below to match the method instead of rewrite it.~~
> disable remapping and use methodname* as method target, with both the mojang mappings and intermediary mappings manually specified
https://github.com/neoforged/NeoForge/blob/821e47fe8e30663d4511530787a16f9d48f38b3f/patches/net/minecraft/client/gui/Gui.java.patch#L221-L226

Update: can't apply solution that mentioned in above, since the signatures of the two methods is different ( (DrawContext) vs (DrawContext, int) ).

Molitvan (approved):
It works

sgmedev (approved):
This works. However, I think it would be better to consider switching to mappings which are consistent such as the offical Mojang mappings or investigate why there is a descrepency in the future instead of duplicating mixins like this.

boholder:
> This works. However, I think it would be better to consider switching to mappings which are consistent such as the offical Mojang mappings or investigate why there is a descrepency in the future instead of duplicating mixins like this.

Indeed. Mojmaps isn't toxic since August 24, 2020. Switching to mojmap means fix all the mismatch of current existing mixins.
https://gist.github.com/boholder/131e5524d7403d48bdcdded5bfd2ddef

BrailleBennett (approved):

PR #430, opened 2025-01-03
Custom Sound playback Logic

### LINKED ISSUE #214: Add sound cues when player's height changes
One for up, one for down.
How to make continuous up and down sounds natural?

### LINKED ISSUE #294: Custom Sound Playing System: Playing extra sound files ported by this mod in built-in position sensitive way
It is more and more important that in-game sound effects can't fulfill this mod's needs.
Need to working on a custom sound playing system, and find some sound effect resources.

https://docs.fabricmc.net/develop/sounds/custom
https://docs.fabricmc.net/develop/sounds/dynamic-sounds
https://docs.fabricmc.net/develop/sounds/using-sounds

rstrunk:
The files relevant to this PR, MainClass and CustomSounds, should now be properly formatted. If this isn't the case, please let me know, and I'll check on another option.

# Changelog

## New Features
* Developers can now make use of the CustomSounds class to introduce sounds from outside of Minecraft. Information on how to do so is in the CustomSounds.java file.


common/src/main/java/org/mcaccess/minecraftaccess/utils/system/CustomSounds.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public static void registerSounds() {
+        for (String soundName : SOUND_NAMES) {
+            try {
+                Identifier temp_id = Identifier.of(MainClass.MOD_ID, soundName);
+                SoundEvent temp_event = SoundEvent.of(temp_id);
+                SOUNDS.register(temp_id, () -> temp_event);
+                REGISTERED_SOUNDS.put(soundName, temp_event);
+                log.info("{} has been loaded.", soundName);
+            } catch (Exception e) {
```
sgmedev: Please never catch `Exception`, this leads to errors being unnoticed and has directly lead to some serious bugs not being caught in the past, it's also generally reguarded as bad practice.

common/src/main/java/org/mcaccess/minecraftaccess/utils/system/CustomSounds.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final String[] SOUND_NAMES = {
+            "y_up",
+            "y_down"
+    };
+
+    public CustomSounds() {
+        try {
+            registerSounds();
+            SOUNDS.register();
+        } catch (Exception E) {
```
sgmedev: Please never catch `Exception`, this leads to errors being unnoticed and has directly lead to some serious bugs not being caught in the past, it's also generally reguarded as bad practice.

common/src/main/java/org/mcaccess/minecraftaccess/utils/system/CustomSounds.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public static void playSoundOnPlayer(String soundName, float volume, float pitch) {
+        SoundEvent soundEvent = REGISTERED_SOUNDS.get(soundName);
+        if (soundEvent == null) {
+            throw new IllegalArgumentException("Not able to find sound: " + soundName);
+        }
+        MinecraftClient.getInstance().player.playSound(soundEvent, volume, pitch);
+    }
+}
```
sgmedev: No newline at end of file, generally it's good practice to end files with a sigular empty line, some tools such as diff (which git uses under the hood) also aren't always super happy when you don't. https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline
```suggestion
}

```

sgmedev (changes requested):
Just a few small things mainly reguarding catching `Exception`, the formatting has definitely improved significantly since last time and is almost perfect now.

With the changlog this is what gets published in the release notes and I'm not sure if what you've got there is overly relevent for users in its current form. I also fail to see where the new custom sounds are actually used, though I havn't had a chance to run it yet because I'm currently away from home and don't have a dev evironment setup on my laptop.

common/src/main/java/org/mcaccess/minecraftaccess/utils/system/CustomSounds.java
```diff
@@ -0,0 +1,61 @@
...
+ * and the sounds themselves should be saved in
+ * "common\src\main\resources\assets\minecraft_access\sounds"
+ * Currently the module only supports playing sounds on the player, but this will change. */
+
+@Slf4j
+public class CustomSounds {
+    public static final Map<String, SoundEvent> REGISTERED_SOUNDS = new HashMap<>();
+    public static final DeferredRegister<SoundEvent> SOUNDS = DeferredRegister.create(MainClass.MOD_ID,
+            RegistryKeys.SOUND_EVENT);
+    private static final String[] SOUND_NAMES = {
```
boholder, line 26: Getting all the sound names from `sounds.json` (with `Gson` or [`Codec`](https://docs.fabricmc.net/develop/codecs)?) instead of hard code them, will reduce redundancy and decouple code from `sounds.json` file.

boholder: Getting all the sound names from `sounds.json` (with `Gson` or [`Codec`](https://docs.fabricmc.net/develop/codecs)?) instead of hard code them, will reduce redundancy and decouple code from `sounds.json` file.

common/src/main/java/org/mcaccess/minecraftaccess/utils/system/CustomSounds.java
```diff
@@ -0,0 +1,61 @@
...
+                SOUNDS.register(temp_id, () -> temp_event);
+                REGISTERED_SOUNDS.put(soundName, temp_event);
+                log.info("{} has been loaded.", soundName);
+            } catch (Exception e) {
+                log.error("Failed to register sound: {}", soundName, e);
+            }
+        }
+    }
+
+    public static void playSoundOnPlayer(String soundName, float volume, float pitch) {
```
boholder: [Here](https://github.com/minecraft-access/minecraft-access/blob/808046b3bda68154f337dd757878c97071de2ef0/common/src/main/java/org/mcaccess/minecraftaccess/utils/PlayerUtils.java#L37-L39) is an existing `playSoundOnPlayer` method. You can modify or extract the inner logic to suit your method's signature.

common/src/main/resources/assets/minecraft_access/sounds/y_down.ogg
boholder, line 1: Sorry but, potential legal issues, I'd like to know how you get these two sounds.

boholder (changes requested):

common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
```diff
@@ -35,7 +36,7 @@ public class MainClass {
     public static PlayerWarnings playerWarnings = null;
     public static AccessMenu accessMenu = null;
     public static FluidDetector fluidDetector = null;
-
+    public static CustomSounds customSounds;
```
boholder: If this class field won't be used anywhere else, you can remove this field and directly invoke the constructor of `CustomSounds`

boholder (changes requested):

common/src/main/java/org/mcaccess/minecraftaccess/utils/system/CustomSounds.java
rstrunk: Agreed. I'll work on this.

common/src/main/java/org/mcaccess/minecraftaccess/utils/system/CustomSounds.java
rstrunk: I initially intended to use that method. Unfortunately, that method takes RegistryEntry.Reference<SoundEvent> as its first arg, where mine takes soundEvent. This is because the method I used to register custom sounds returns a sound event directly, where the one in PlayerUtils does what works well for vanilla sounds. It's possible we could create an overloaded method in PlayerUtils that calls a different minecraft method - like mine does. Thoughts?

common/src/main/resources/assets/minecraft_access/sounds/y_down.ogg
rstrunk: All sounds I include will either be royalty-free or created by me. In the case of these two, I pitched up a tone generator for the up, then reversed the sound for the down.

common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
rstrunk: I can't tell which field you're referring to.

common/src/main/java/org/mcaccess/minecraftaccess/utils/system/CustomSounds.java
rstrunk: Will fix that. Thanks.

common/src/main/java/org/mcaccess/minecraftaccess/utils/system/CustomSounds.java
boholder: > create an overloaded method in PlayerUtils that calls a different minecraft method - like mine does.

Yeah, that's exactly what I mean

common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
boholder: > which field you're referring to.

`public static CustomSounds customSounds;`

boholder (approved):

sgmedev (changes requested):
I still fail to see where the new custom sounds are actually used and therefore am unable to test it. Merging unused code can be risky because if the followup PR ends up getting delayed for any reason, we could end up with vestigial code which someone has to maintain.

I am also not a fan of loading everything into a HashMap because this lacks any form of compile-time checking, could something akin to Minecraft's `SoundEvents` class be implemented instead?

boholder:
> I still fail to see where the new custom sounds are actually used and therefore am unable to test it.

The following logic is in #432. @rstrunk  How about moving the Y coordination logic from #432 to this PR?

rstrunk:
@TheSuperGamer20578 , can you please talk me through your hashmap concern? I'm using that to hold the sound_events and their associated names. Minecraft uses the registry for the same purpose, but I couldn't find a way to add to a registry with Architecture.
As for the sounds themselves, it is possible to test to see if it works, but it requires modifying source a little bit. I'm not sure how to demo the functionality without introducing something unrelated to playing custom sounds.

sgmedev:
The problem with a HashMap is that there are absolutely no compile time checks and we don’t get any IDE autocompletion. This means that in order to use any of the sounds, we have to manually refer to the JSON file, and if we typo the sound name, it won’t work, and there won’t be any error until it attempts to play the sound, adding yet another thing that needs to be manually checked. The Fabric docs have a page on creating custom sounds with an example on how to store the sounds as constants https://docs.fabricmc.net/develop/sounds/custom, this among other pages was linked in #294.

As for custom sounds being unused in this PR, being unable to test it is only part of the problem. The main problem is that merging unused code bears some risk, should your follow-up PR be significantly delayed for any reason, we will be stuck maintaining this unused code. We have faced this problem before already: on the 5th of September 2023 the first PR in a series of PRs for the area map (#154) was merged, just over a month later, the PRs for this feature stopped until a few days ago when we decided to remove it and put it on its own branch in #433. In the meanwhile, for almost a year and a half, half our test suite was for this vestigial code and had to be maintained, and translators were translating the non-insignificant number of new strings. While this PR is nowhere near the scale of the area map, we still don’t want to risk having to maintain any vestigial code again.

rstrunk:
@boholder
What if, instead of putting in any of the coordinate changes code, I switch the sound of an existing feature from something Vanilla to something custom?
The coordinate changes sounds are optional and might annoy some people. If I built something in here, I would have to write a config for it as well. If I just change something existing to a custom sound, the structure will already be in place.

rstrunk:
@TheSuperGamer20578,
I originally had a list of constants in the first PR I put up based on the Fabric documentation you referenced above, but I took it out because I was told it required maintaining both the list in the source code and the json file. Your point about IDE completion is a good one. Can you please recommend an approach for me to explore?

sgmedev:
I think doing what the Fabric Wiki suggests and having each `SoundEvent` be its own constant as seen below is probably the best idea. If we wanted to not have to maintain a `sounds.json` seperately we could probably generate it with an annotation processor, but this would likely be complex and would fit better in a library.
```java
public static final SoundEvent ITEM_METAL_WHISTLE = registerSound("metal_whistle");
public static final SoundEvent ENGINE_LOOP = registerSound("engine");
```

rstrunk:
@TheSuperGamer20578,
I revised it because of this request two weeks ago.
"Getting all the sound names from sounds.json (with Gson or [Codec](https://docs.fabricmc.net/develop/codecs)?) instead of hard code them, will reduce redundancy and decouple code from sounds.json file."
Are we moving away from that recommendation now?

sgmedev:
@boholder requested that not me so they would probably be better to answer that question but I would say yes due to reasons stated previously.

boholder:
@rstrunk I still suggest adding the Y coordination logic to this PR, for config code concerns, maybe you can do it after #347 is merged, or you can simply add a boolean config field to control the feature, without config menu.

rstrunk:
@boholder  and what about the above regarding loading from json versus having the sounds directly in the code so they can be referenced?

boholder:
> @boholder  and what about the above regarding loading from json versus having the sounds directly in the code so they can be referenced?

The key concept is don't duplicate, whether maintaining the sounds info in json or code doesn't matter... to me at least. So, json or code? I haven't look into the details thus I can't tell, sorry...

rstrunk:
@boholder  and @TheSuperGamer20578 ,
One more thought that might solve both your needs, but I am really not excited about the possibility.
The Json file is required in order to load the sounds, but TSG understandably wants the variables referenced in the code so the IDE can see them. Would it be possible to create the Json file from code before the sounds are registered, or would Mc not see it, then?
If it worked, would that be a suitable solution?

sgmedev:
Resources have to be added at compile-time, so it’s not as simple as just creating a JSON file at runtime. You could probably do something crazy and insane with annotation processors or a Gradle script or plugin to create the JSON file or Java source, but doing that in this codebase would create way more complexity than it saves.

boholder:
@TheSuperGamer20578 So... what's the current best solution? Sort of confused

rstrunk:
@TheSuperGamer20578  and @boholder ,
Since the code is no longer valid because mappings changed, and since we don't have a consensus on the above, I recommend we just close this.

Issue #391, opened 2024-11-16
Variables for things like biome and xp aren't cleared on world switch


BrailleBennett:
When you switch worlds/servers in the same session, things like the previous biome and xp aren't cleared, so if you leave a world in a planes biome and join a new one in a plains biome the biome entered message won't be spoken.

PR #436, opened 2025-01-12
Add effect narration

### LINKED ISSUE #217: Add Effect narration
https://minecraft.wiki/w/Effect
Such as notifying remain time...

Molitvan:
Closes #217 

# Changelog

## New Features

- New keybinding `Control` + `Speak Player Status`: Speak currently active effects
- Automatically speak gained and lost effects (when configuration option `Enable Player Status` is `on`)


BrailleBennett (dismissed):
You should take into account the discussion in:
https://github.com/minecraft-access/minecraft-access/issues/217#issuecomment-2560889962

Molitvan:
> You should take into account the discussion in: [#217 (comment)](https://github.com/minecraft-access/minecraft-access/issues/217#issuecomment-2560889962)

After reading that, I think the only thing I should add is narrating amplifiers (I forgot to do that). I think dividing into buffs and debuffs is not needed and it would need to be done manually unless I'm missing something, and also I'm not sure how to know that for example Haste is from a beacon.

The current narration looks something like:
- When you press X: Night Vision 1:13, Speed 2 4:32, Blindness ∞
- When gaining an effect: Gained Jump Boost 2
- When losing an effect: Lost Night Vision

BrailleBennett:
You should maybe also narrate whether the effect came from a beacon as this is shown by a blue outline for sighted players

Molitvan:
> You should maybe also narrate whether the effect came from a beacon as this is shown by a blue outline for sighted players

After researching, I found that the blue outline does not mean the effect came from a beacon, but that it is ambient aka that there are no particles. From what I understand, without commands the only ways to get an ambient effect is either by a beacon or a conduit. There is StatusEffectInstance#isAmbient so I can very easily check for this, but I don't know how to narrate it and honestly I don't think that information is that useful.

boholder:
Maybe appending the effect narration after player status is better, assigning precious single key to this feature is... not properly, IMO, sorry.

common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
```diff
@@ -148,6 +148,8 @@ private static void _clientTickEventsMethod(MinecraftClient minecraftClient) {
         FallDetector.getInstance().update();
 
         Keystroke.updateInstances();
+
+        EffectNarration.getInstance().update();
```
boholder: Well, although the new `EffectNarration` feature hasn't used custom `Keystroke` classes, it's better to move this line before `Keystroke.updateInstances();`

common/src/main/java/org/mcaccess/minecraftaccess/features/EffectNarration.java
```diff
@@ -0,0 +1,93 @@
...
+public class EffectNarration {
+    @Getter
+    private static EffectNarration instance = new EffectNarration();
+
+    private List<StatusEffectInstance> previousEffects = new ArrayList<>();
+
+    public void update() {
+        if (MinecraftClient.getInstance() == null) return;
+        if (MinecraftClient.getInstance().currentScreen != null) return;
+        if (MinecraftClient.getInstance().player == null) return;
```
boholder: Or you can move the feature update line in `MainClass` inside the null check scope, so you can remove these null checks.
https://github.com/minecraft-access/minecraft-access/blob/163683af5a8a634a7bdcc7da803e61374f43a4ee/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java#L126-L137

common/src/main/java/org/mcaccess/minecraftaccess/features/EffectNarration.java
```diff
@@ -0,0 +1,93 @@
...
+
+        if (KeyUtils.isAnyPressed(KeyBindingsHandler.getInstance().speakEffectsKey)) {
```
boholder: That's why we introduce `Keystroke`, you may not noticed, if you continue pressing the key, the feature will spam in the debug log.
https://github.com/minecraft-access/minecraft-access/blob/8a1103b43ed9b7d4ec34972a7a41602ca86383ed/common/src/main/java/org/mcaccess/minecraftaccess/utils/condition/Keystroke.java#L13-L17

common/src/main/java/org/mcaccess/minecraftaccess/features/EffectNarration.java
```diff
@@ -0,0 +1,93 @@
...
+            if (MinecraftClient.getInstance().player.getStatusEffects().isEmpty()) {
+                MainClass.speakWithNarrator(I18n.translate("minecraft_access.effect_narration.no_effects"), true);
+                return;
+            }
+
+            StringBuilder toSpeak = new StringBuilder();
+            for (StatusEffectInstance effect : MinecraftClient.getInstance().player.getStatusEffects()) {
```
boholder: Please make use of util methods, replace `MinecraftClient.getInstance().player` with `WorldUtils.getClientPlayer()` (and the same code in rest of pr)

common/src/main/java/org/mcaccess/minecraftaccess/features/EffectNarration.java
```diff
@@ -0,0 +1,93 @@
...
+                toSpeak.append(NarrationUtils.narrateEffect(effect));
+            }
+
+            MainClass.speakWithNarrator(toSpeak.toString(), true);
+        }
+
+        List<StatusEffectInstance> currentEffects = new ArrayList<>(MinecraftClient.getInstance().player.getStatusEffects());
```
boholder: It would be better if we can find out the proper injection points for gaining and losing effect, and put the narration logic with mixin, doing so to save a big chunk of code here. Another reason why we should try to avoiding cache states with `previous*` class field is it will cause issues like #391

common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
```diff
@@ -568,4 +569,27 @@ private static String getFluidI18NName(RegistryEntry<Fluid> fluid) {
         );
         return I18n.translate(translationKey);
     }
+
+        public static String narrateEffect(StatusEffectInstance effect) {
+        StringBuilder result = new StringBuilder();
```
boholder, line 568: Wrong indentation?

common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
```diff
@@ -568,4 +569,27 @@ private static String getFluidI18NName(RegistryEntry<Fluid> fluid) {
...
+
+        int amplifier = effect.getAmplifier();
+        if (amplifier > 1) {
+            result.append(String.valueOf(amplifier)).append(" ");
+        }
+
+        if (effect.isInfinite()) {
+            result.append(I18n.translate("effect.duration.infinite"));
+        } else {
+            int duration = effect.getDuration() / 20;
```
boholder: Please leave a comment about why divide by 20.

common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
```diff
@@ -568,4 +569,27 @@ private static String getFluidI18NName(RegistryEntry<Fluid> fluid) {
...
+            int minutes = duration / 60;
```
boholder: BTW, you can use:

```java
import org.apache.commons.lang3.time.DurationFormatUtils;

Duration d = Duration.ofSeconds(70);
String fmt = d.toHoursPart() == 0? "mm':'ss" : "HH':'mm':'ss";
// 1:10
String toSpeak = DurationFormatUtils.formatDuration(d.toMillis(), fmt);
```

- https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/Duration.html
- https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/time/DurationFormatUtils.html

boholder (dismissed):

BrailleBennett:
Make sure to edit your changelog

Molitvan:
> That's why we introduce Keystroke, you may not noticed, if you continue pressing the key, the feature will spam in the debug log.

I'll keep that in mind for future PRs. It's not a problem here anymore since I removed the separate keybind.

> It would be better if we can find out the proper injection points for gaining and losing effect, and put the narration logic with mixin, doing so to save a big chunk of code here. Another reason why we should try to avoiding cache states with previous* class field is it will cause issues like #391

You're right. I didn't think about that. I propose we reinitialize all feature classes on world join to handle things that we can't find good injection points for. I'm not sure where to inject for this without the resulting narration looking like spam though.


>BTW, you can use:
> ```java
import org.apache.commons.lang3.time.DurationFormatUtils;

Duration d = Duration.ofSeconds(70);
String fmt = d.toHoursPart() == 0? "mm':'ss" : "HH':'mm':'ss";
// 1:10
String toSpeak = DurationFormatUtils.formatDuration(d.toMillis(), fmt);
> ```
> • https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/Duration.html
> • https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/time/DurationFormatUtils.html

This seems like overcomplicating a bit.

boholder:
> This seems like overcomplicating a bit.

I disagree with it, a bit more simpler

common/src/main/java/org/mcaccess/minecraftaccess/features/EffectNarration.java
```diff
@@ -0,0 +1,86 @@
...
+            boolean isNewEffect = true;
+            for (StatusEffectInstance previousEffect : previousEffects) {
+                if (effect.getEffectType().equals(previousEffect.getEffectType())) {
+                    isNewEffect = false;
+                    break;
+                }
+            }
+            if (isNewEffect) {
+                newEffects.add(effect);
+            }
```
sgmedev: Please use the contains method.

common/src/main/java/org/mcaccess/minecraftaccess/features/EffectNarration.java
```diff
@@ -0,0 +1,86 @@
...
+            boolean isLostEffect = true;
+            for (StatusEffectInstance currentEffect : currentEffects) {
+                if (effect.getEffectType().equals(currentEffect.getEffectType())) {
+                    isLostEffect = false;
+                    break;
+                }
+            }
+            if (isLostEffect) {
+                lostEffects.add(effect);
+            }
```
sgmedev: Please use the contains method.

common/src/main/java/org/mcaccess/minecraftaccess/features/EffectNarration.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        StringBuilder toSpeak = new StringBuilder();
+
+        if (!newEffects.isEmpty()) {
+            toSpeak.append(I18n.translate("minecraft_access.effect_narration.gained")).append(" ");
+            for (StatusEffectInstance effect : newEffects) {
+                toSpeak.append(I18n.translate(effect.getTranslationKey())).append(", ");
+            }
```
sgmedev: This could be simplified by using Collectors.joining() on a stream, this would also avoid the trailing comma.

common/src/main/java/org/mcaccess/minecraftaccess/features/EffectNarration.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        if (!lostEffects.isEmpty()) {
+            toSpeak.append(I18n.translate("minecraft_access.effect_narration.lost")).append(" ");
+            for (StatusEffectInstance effect : lostEffects) {
+                toSpeak.append(I18n.translate(effect.getTranslationKey())).append(", ");
+            }
```
sgmedev: This could be simplified by using Collectors.joining() on a stream, this would also avoid the trailing comma.

common/src/main/java/org/mcaccess/minecraftaccess/features/EffectNarration.java
```diff
@@ -0,0 +1,86 @@
...
+import net.minecraft.client.resource.language.I18n;
+import net.minecraft.entity.effect.StatusEffectInstance;
+
+public class EffectNarration {
+    @Getter
+    private static EffectNarration instance = new EffectNarration();
+
+    private List<StatusEffectInstance> previousEffects = new ArrayList<>();
+
+    public void update() {
```
sgmedev: This whole thing can be simplified by using Sets. gained = new - old, lost = old - new.

Even better would be to add an event handler (if such an event exists) or inject into where effects are added or removed.

sgmedev (changes requested):

sgmedev (dismissed):
A few slight issues:
- When gaining multiple effects at the same time (e.g. by eating a golden apple or enchanted golden apple) only the last effect is read
- Effects keep getting read periodically
- Beacon effects get read every ten seconds

boholder:
>     * When gaining multiple effects at the same time (e.g. by eating a golden apple or enchanted golden apple) only the last effect is read

That's how mixin solution works, if we switch back to client tick event solution, we still suffer this if the effects aren't applied within same tick.
I want to but there is no way to control the screen reader to not suppress narrations.
There is a simple fix, just change the override param of `speak` method to false...

>     * Effects keep getting read periodically
>     * Beacon effects get read every ten seconds

It's because the narration is triggered when effects are updated (time, level).
Fixing this requires introducing a cache for comparing narration, and an interval config to indicate suppression duration of same narration.
But yeah, this is a problem, I'll fix it.

boholder (approved):

PR #439, opened 2025-01-13
Add title narration

### LINKED ISSUE #403: Add support for titles
Currently, the mod doesn't narrate titles. In vanilla, titles are usually created with the /title command and they are widely used across maps and servers.

Note: the actionbar that the /title command refers to is not the same as the actionbar mentioned in #402. In vanilla, that actionbar is only used to display messages like "You can only sleep at night!" and to my knowledge, there is no way to display your own messages in there without mods or plugins.

Molitvan:
Closes #403 

# Changelog

## New Features
- Added narration for titles and subtitles

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/mixin/InGameHudMixin.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/mixin/InGameHudMixin.java
@@ -31,6 +31,12 @@ public class InGameHudMixin {
     @Shadow
     private ItemStack currentStack;
 
+    @Shadow
+    private Text title;
+
+    @Shadow
+    private Text subtitle;
+
     @Unique
     private final SpeakHeldItem minecraft_access$feature = new SpeakHeldItem();
 
@@ -71,11 +77,6 @@ public class InGameHudMixin {
         String toSpeak = String.join(", ", parts);
         MainClass.speakWithNarratorIfNotEmpty(toSpeak, true);
     }
-
-    @Shadow
-    private Text title;
-    @Shadow
-    private Text subtitle;
 
     @Inject(method = "setTitle", at = @At("TAIL"))
     public void setTitleMixin(Text title, CallbackInfo ci) {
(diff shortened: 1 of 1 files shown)


sgmedev (changes requested):
Subtitles still aren't read, otherwise this works perfectly.

Molitvan:
> Subtitles still aren't read, otherwise this works perfectly.

My bad. Apparently subtitles have their own method. One sec

sgmedev:
One thing to watch out for is it appears that the subtitle only displays when the title displays, at least thats how it worked with my limited testing with the `/title` command.

Molitvan:
> One thing to watch out for is it appears that the subtitle only displays when the title displays, at least thats how it worked with my limited testing with the `/title` command.

That's right. I accounted for that.

common/src/main/java/org/mcaccess/minecraftaccess/mixin/InGameHudMixin.java
```diff
@@ -70,4 +71,17 @@ public void speakActionbar(Text message, boolean tinted, CallbackInfo ci) {
...
         MainClass.speakWithNarratorIfNotEmpty(toSpeak, true);
     }
+
+    @Inject(method = "setTitle", at = @At("TAIL"))
+    public void setTitleMixin(Text title, CallbackInfo ci) {
+        MainClass.speakWithNarrator(title.getString(), true);
+    }
+
+    @Shadow
+    private Text title;
```
sgmedev: Fields should come before methods.

sgmedev (changes requested):
Almost perfect, only problem is that if I set a subtitle before a title the subtitle is completely ignored, it should still be read when the title appears though.

Molitvan:
Hope I got everything right this time. I was testing like this:
1. An impulse command block setting the title and a chain command block setting the subtitle
2. An impulse command block setting the subtitle and a chain command block setting the title

sgmedev (approved):

BrailleBennett (approved):

PR #442, opened 2025-01-16
Made toasts narratable

### LINKED ISSUE #408: Read toasts
Currently toasts (the notifications that appear in the top-right) are not read at all. It should be possible to either bind an event handler or mixin into whatever is responsible for displaying toasts.

BrailleBennett:

# Changelog

## New Features
- Made toasts (advancement, recipe unlock, system, and tutorial messages) narrated by the mod


common/src/main/java/org/mcaccess/minecraftaccess/mixin/ToastManagerMixin.java
```diff
@@ -0,0 +1,60 @@
...
+import org.mcaccess.minecraftaccess.MainClass;
+import org.mcaccess.minecraftaccess.utils.StringUtils;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.util.stream.Collectors;
+
+@Slf4j
```
sgmedev: Unused
```suggestion
```

sgmedev (dismissed):

BrailleBennett:
Merge #443 before this please.

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ToastManagerMixin.java
```diff
@@ -0,0 +1,59 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.util.stream.Collectors;
+
+@Mixin(ToastManager.class)
+public class ToastManagerMixin {
+    @Inject(at = @At("TAIL"), method = "add")
+    public void add(Toast toast, CallbackInfo ci) {
```
boholder: Suggest changing the `add` method name to something like `narrateToast`

common/src/main/java/org/mcaccess/minecraftaccess/utils/StringUtils.java
```diff
@@ -42,4 +49,14 @@ public static String[] splitToParts(String msg) {
             return msg.split("\\s");
         }
     }
+
+    @Contract(pure = true)
+    public static @NotNull String orderedTextToString(@NotNull OrderedText text) {
+        StringBuilder builder = new StringBuilder();
+        text.accept((index, style, codePoint) -> {
+            builder.appendCodePoint(codePoint);
```
boholder, line 54: Raw codepoint processing isn't a good idea. After digging the code, `SystemToast.getTextAsList` ->`Text.asOrderedText` -> `MutableText.asOrderedText` ... end up with `ReorderingUtil`. It seems like the `OrderedText` is for drawing the text on the screen, for example"reverting" Arabic texts since they are shown in the top-down order. For me, check English, Chinese, Arabic three languages is enough for ensure the result of `builder.toString()` is always right, but, who can read Arabic text?

https://stackoverflow.com/questions/23979676/java-what-are-characters-code-points-and-surrogates-what-difference-is-there

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ToastManagerMixin.java
```diff
@@ -0,0 +1,59 @@
...
+        StringBuilder toastTextBuilder = new StringBuilder();
+        toastTextBuilder.append(I18n.translate("minecraft_access.toast.shown"))
+                .append(", ");
+        switch (toast) {
+            case AdvancementToast advancementToast -> {
```
boholder: brackets behind each `case * ->` can be omitted

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ToastManagerMixin.java
```diff
@@ -0,0 +1,59 @@
...
+                ((AdvancementToastAccessor) advancementToast).getAdvancement()
+                        .value().display()
+                        .ifPresent(display -> {
+                            toastTextBuilder.append(display.getFrame().getToastText().getString())
+                                    .append(' ')
```
boholder: There used to be a `separator` i18n text and now disappears, In Chinese, the narration is "Toast shown, 新配方已解锁！. 请检查你的配方书", there is a the unnatural point behind exclamation.

boholder (dismissed):
Checked with Chinese language setting.

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ToastManagerMixin.java
BrailleBennett: This is done now.

common/src/main/java/org/mcaccess/minecraftaccess/utils/StringUtils.java
BrailleBennett: It's definitely not a clean solution, it's just better than the alternatives unless I'm missing something

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ToastManagerMixin.java
boholder: @BrailleBennett this is leftover

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ToastManagerMixin.java
boholder: @BrailleBennett Have you found the corresponding original i18n text?

boholder (approved):

sgmedev (approved):

PR #450, opened 2025-02-01
Fix narration of numbers > 10M

BrailleBennett:

# Changelog

## Bug Fixes
- If a number was at least plus or minus 10 million (like coordinates), it was improperly truncated

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/features/PositionNarrator.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/features/PositionNarrator.java
@@ -33,6 +33,9 @@ public class PositionNarrator {
         positionNarrationKey = new Keystroke(() -> KeyUtils.isAnyPressed(KeyBindingsHandler.getInstance().positionNarrationKey));
     }
 
+    private PositionNarrator() {
+    }
+
     public void update() {
         if (!OtherConfigsMap.getInstance().isPositionNarratorEnabled()) return;
 
--- a/common/src/main/java/org/mcaccess/minecraftaccess/features/PositionNarrator.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/features/PositionNarrator.java
@@ -33,37 +33,27 @@ public class PositionNarrator {
         positionNarrationKey = new Keystroke(() -> KeyUtils.isAnyPressed(KeyBindingsHandler.getInstance().positionNarrationKey));
     }
 
-    private PositionNarrator() {
-    }
-
     public void update() {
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/mcaccess/minecraftaccess/utils/position/PlayerPositionUtils.java
```diff
@@ -18,31 +18,18 @@ public class PlayerPositionUtils {
...
-
-        return Double.parseDouble(tempPosZ);
+        return Math.round(getPlayerPosition().orElseThrow().z * 10.0) / 10.0;
     }
 
-    /**
-     * Player position is at player's leg.
-     */
+     // Player position is at player's leg.
     public static Optional<Vec3> getPlayerPosition() {
```
boholder, line 38: Java doc is better than double slash comment, for method.

common/src/main/java/org/mcaccess/minecraftaccess/utils/position/PlayerPositionUtils.java
```diff
@@ -18,31 +18,18 @@ public class PlayerPositionUtils {
     private static final String POSITION_FORMAT = "{x}, {y}, {z}";
 
     public static double getX() {
-        String tempPosX = String.valueOf(getPlayerPosition().orElseThrow().x);
-        tempPosX = tempPosX.substring(0, tempPosX.indexOf(".") + 2);
-
-        return Double.parseDouble(tempPosX);
+        return Math.round(getPlayerPosition().orElseThrow().x * 10.0) / 10.0;
```
boholder, line 22: The original logic is round to two bits while the new logic is round to one bit, I prefer the new one, one bit is enough.

boholder (dismissed):

common/src/main/java/org/mcaccess/minecraftaccess/features/PositionNarrator.java
```diff
@@ -33,37 +33,27 @@ public class PositionNarrator {
         positionNarrationKey = new Keystroke(() -> KeyUtils.isAnyPressed(KeyBindingsHandler.getInstance().positionNarrationKey));
     }
 
-    private PositionNarrator() {
-    }
-
```
sgmedev: This should stay private.

sgmedev (changes requested):

sgmedev (dismissed):

boholder (approved):

sgmedev (approved):

PR #463, opened 2025-02-11
Made more numbers go through the narration util

### LINKED ISSUE #458: Avoid rounding coordinates when output isn't spoken
Right now, coordinates are rounded when the mod gets them from our built in utility methods to the nearest tenth of a decimal. We should avoid rounding either by adding a parameter Boolean to say should round or by only performing the rounding when the number is about to be spoken allowed, perhaps in the speakNumber method where the "negative" prefix is applied now?

BrailleBennett:

# Changelog

## Feature Updates
- If numbers ended in .0, they will now be trimmed to exclude it
## Bug Fixes
- All spoken numbers now correctly only narrate out to the first number after the decimal point if they are a decimal
- 12:00 AM is now correctly narrated when listening to the time with 12 hour format instead of 00:00 AM


boholder:
Isn't "10ths place" too precise?... And I can't find the related code for "10ths place", cloud you point it out?

common/src/main/java/org/mcaccess/minecraftaccess/features/SpeakHeldItem.java
```diff
@@ -37,7 +38,7 @@ private void speakIfHeldChanged(String itemName, int itemCount, Function<String,
         boolean reportHeldItemsCountWhenChanged = OtherConfigsMap.getInstance().isReportHeldItemsCountWhenChanged();
 
         if (nameChanged) {
-            String itemCountText = itemCount == 0 ? "" : itemCount + " ";
+            String itemCountText = itemCount == 0 ? "" : NarrationUtils.narrateNumber(itemCount) + " ";
```
boholder, line 45: Some numbers will never be float type, like this one.

common/src/main/java/org/mcaccess/minecraftaccess/features/XPIndicator.java
```diff
@@ -34,8 +35,8 @@ public void update() {
         boolean increased = previousXPLevel < currentXPLevel;
         previousXPLevel = currentXPLevel;
 
-        String toSpeak = (increased) ? I18n.get("minecraft_access.xp_indicator.increased", currentXPLevel)
-                : I18n.get("minecraft_access.xp_indicator.decreased", currentXPLevel);
+        String toSpeak = (increased) ? I18n.get("minecraft_access.xp_indicator.increased", NarrationUtils.narrateNumber(currentXPLevel))
```
boholder, line 38: And this one

common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
```diff
@@ -154,9 +154,14 @@ private static String getSheepInfo(Sheep sheepEntity, String currentQuery) {
         return translatedColor + " " + shearable;
     }
 
-    public static String narrateNumber(double d) {
+    public static String narrateNumber(double num) {
         DecimalFormat df = new DecimalFormat();
-        return d >= 0 ? String.valueOf(df.format(d)) : I18n.get("minecraft_access.other.negative", df.format(-d));
+        num = Math.round(num * 10.0) / 10.0;
```
boholder, line 159: By "10ths place" you mean this?

boholder (dismissed):

boholder (dismissed):
The change log, "only narrate out to the 10ths place" might be changed to "narrated to one decimal place" to make it more clearer

common/src/main/java/org/mcaccess/minecraftaccess/features/access_menu/AccessMenu.java
```diff
@@ -141,119 +137,88 @@ private static void handleInMenuActions() {
...
-            long daytime = minecraftClient.player.clientLevel.getDayTime() + 6000;
-            int hours = (int) (daytime / 1000) % 24;
-            int minutes = (int) ((daytime % 1000) * 60 / 1000);
-
-            String translationKey = "minecraft_access.access_menu.time_of_day";
-            if (OtherConfigsMap.getInstance().isUse12HourTimeFormat()) {
-                if (hours > 12) {
-                    hours -= 12;
-                    translationKey += "_pm";
-                } else if (hours == 12) {
```
sgmedev, line 245: This line should not be removed. Removing this line causes noon to be read as `12:00 AM` and causes any later time to be untranslated as `minecraft_access.access_menu.time_of_day_pm_pm`.

sgmedev (changes requested):

common/src/main/java/org/mcaccess/minecraftaccess/features/access_menu/AccessMenu.java
BrailleBennett: Super good catch, I think I accidentally removed it, but it helped me find another bug with midnight as well. So it was a win win.

boholder (dismissed):

boholder (approved):

sgmedev (approved):

PR #465, opened 2025-02-13
feat: remove formatting codes in narration

boholder:
BTW, the Narrator (ScreenReader[Windows/MacOS/Linux]) implementation does need a refactor or rewrite to make it easy to understand #475 

# Changelog

## Feature Updates

- Remove [formatting codes](https://minecraft.wiki/w/Formatting_codes) from all narrations


BrailleBennett:
Are you referring to
https://minecraft.wiki/w/Formatting_codes#:~:text=This%20article%20is%20about%20the,be%20removed%20in%20the%20future.
Do we know if there is a deprication timeline?

boholder:
> deprication

This patch works for current version. Don't know anything else.

BrailleBennett (dismissed):

common/src/main/java/org/mcaccess/minecraftaccess/mixin/NarrationThunkMixin.java
```diff
@@ -0,0 +1,32 @@
...
+    private T contents;
+
+    /**
+     * <a href="https://minecraft.wiki/w/Formatting_codes">wiki page of formatting codes</a>
+     */
+    @SuppressWarnings("unchecked")
+    @Inject(at = @At("TAIL"), method = "<init>")
+    private void removeFormattingCodes(Object value, BiConsumer<Consumer<String>, T> transformer, CallbackInfo ci) {
+        if (value instanceof String) {
+            this.contents = (T) ((String) value).replaceAll("§.", "");
```
sgmedev: A pattern variable should be used here isntead of a manual cast. The `this.` is also redundent.
```suggestion
        if (value instanceof String string) {
            contents = (T) string.replaceAll("§.", "");
```

common/src/main/java/org/mcaccess/minecraftaccess/mixin/NarrationThunkMixin.java
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+}
```
sgmedev: No newline at end of file.

common/src/main/java/org/mcaccess/minecraftaccess/mixin/NarrationThunkMixin.java
```diff
@@ -0,0 +1,32 @@
...
+    @Final
+    @Shadow
+    private T contents;
+
+    /**
+     * <a href="https://minecraft.wiki/w/Formatting_codes">wiki page of formatting codes</a>
+     */
+    @SuppressWarnings("unchecked")
+    @Inject(at = @At("TAIL"), method = "<init>")
+    private void removeFormattingCodes(Object value, BiConsumer<Consumer<String>, T> transformer, CallbackInfo ci) {
```
sgmedev: Shouldn't the type of `value` be `T`?
```suggestion
    private void removeFormattingCodes(T value, BiConsumer<Consumer<String>, T> transformer, CallbackInfo ci) {
```

sgmedev (changes requested):
This does not appear to work. When I test it by running `tellraw @a "§ltest"` in the server console it still says `§ltest` instead of just `test`.

boholder:
> This does not appear to work. When I test it by running `tellraw @a "§ltest"` in the server console it still says `§ltest` instead of just `test`.

Mine game forced the player logout after pressing enter key to send:

```log
[13:46:10] [Render thread/INFO] (ScreenReaderLinux) Speaking(interrupt:true)= Chat screen. /. Input message and press Enter to send
[13:46:11] [Render thread/INFO] (ScreenReaderLinux) Speaking(interrupt:true)= /tellraw @a "§ltest"
[13:46:12] [Server thread/INFO] (Minecraft) Player330 lost connection: Illegal characters in chat
[13:46:12] [Server thread/INFO] (Minecraft) Player330 left the game
[13:46:12] [Render thread/INFO] (ScreenReaderLinux) Speaking(interrupt:true)= Connection Lost. Illegal characters in chat
[13:46:12] [Server thread/INFO] (Minecraft) Stopping singleplayer server as player logged out
```

sgmedev:
I ran it from the server console, try that.

boholder:
> I ran it from the server console, try that.

Confirmed

sgmedev (dismissed):
I'm still experiancing the same issue. When I press `alt`+`1` with the chat open it works, but not when the message is recieved and read automatically. I do really like the simplicity of this solution though.

boholder:
> but not when the message is recieved and read automatically

Ah, I get, the chat narrating-on-receive logic must be an original one, I'll make the interception position deeper into the narrator api.

sgmedev (dismissed):

sgmedev (dismissed):

BrailleBennett (dismissed):

sgmedev (approved):

BrailleBennett (approved):

PR #468, opened 2025-02-18
Removed all refrences to access wideners

BrailleBennett:
This removes the line in build.gradle that was attempting to package unused access widener files into the mod jar.

PR #471, opened 2025-02-24
Switch to gradle versions catalog and add dependabot management

BrailleBennett:
This PR switches to the [gradle versions catalog](https://docs.gradle.org/current/userguide/version_catalogs.html) system. This switch not only centralizes everything version controlled across the gradle project into one easy to understand file, but it also enables us to leverage GitHub's powerful dependency management tool to insure we are always up to date with predictable and well formatted PRs that we can choose to test and merge ourselves. In addition to gradle management, I also enabled GitHub Actions management, so all components in our workflows can also be managed by dependabot. Currently dependabot is set up to:
- Perform submodule updates daily
- Perform GitHub Actions pull requests weekly (only minor and patch updates)
- Perform gradle version catalog pull requests weekly (only minor and patch updates)

I chose to leave off the major updates, as things are planned to break when projects switch major versions, so it gives us the free will to test and update if we want. This does mean that that aspect is still manual, but it is by design. Sadly, some projects like Minecraft themselves do NOT follow semver guidelines, so when a Minecraft update happens, we will have to be aware and most likely hold off on updating dependencies until the mod is ready to update.
You can find an example of how a pull request might look here:
https://github.com/BrailleBennett/minecraft-access/pull/17


fabric/build.gradle
```diff
@@ -31,10 +28,9 @@ configurations {
 }
 
 dependencies {
-    modImplementation "net.fabricmc:fabric-loader:${rootProject.fabric_loader_version}"
-    modImplementation include("net.fabricmc.fabric-api:fabric-api:${project.fabric_api_version}")
-
-    modImplementation include("dev.architectury:architectury-fabric:${project.architectury_api_version}")
+    modImplementation libs.fabricLoader
+    modImplementation include(libs.fabricApi.get().toString())
+    modImplementation include(libs.architecturyApiFabric.get().toString())
```
sgmedev: This `.toString()` is quite messy. Do we even directly depend on either of these in the `fabric` subproject? If not we should be able to completely remove the `modImplementation` and just use `include`.

neoforge/build.gradle
```diff
@@ -30,37 +27,24 @@ configurations {
...
-        url = 'https://maven.neoforged.net/releases'
-    }
-}
-
 dependencies {
-    neoForge "net.neoforged:neoforge:${rootProject.neoforge_version}"
+    neoForge libs.neoForge
 
-    modImplementation include("dev.architectury:architectury-neoforge:${project.architectury_api_version}")
+    modImplementation include(libs.architecturyApiNeoForge.get().toString())
```
sgmedev: This `.toString()` is quite messy. Do we even directly depend on this in the `neoforge` subproject? If not we should be able to completely remove the `modImplementation` and just use include.

settings.gradle
```diff
@@ -1,12 +1,54 @@
...
+        }
+        maven {
+            name = 'Neoforged'
+            url = 'https://maven.neoforged.net/releases/'
+        }
+        maven {
+            name = 'ParchmentMC'
+            url = 'https://maven.parchmentmc.org'
+        }
+    }
```
sgmedev: What order is this in?

build.gradle
```diff
@@ -23,50 +19,61 @@ if (System.getenv("GITHUB_ACTIONS") == "true") {
...
-
-allprojects {
-    apply plugin: "java"
-    apply plugin: "architectury-plugin"
-    apply plugin: "maven-publish"
-
-    archivesBaseName = rootProject.archives_base_name
-    group = rootProject.maven_group
 
+//do NOT add anything to this block
```
sgmedev:
```suggestion
// Do NOT add anything to this block
```

sgmedev (changes requested):

.github/dependabot.yml
```diff
@@ -4,3 +4,25 @@ updates:
...
+          - patch
+
+  - package-ecosystem: gradle
+    directory: "/"
+    schedule:
+      interval: weekly
+    groups:
+      gradle-dependencies:
+        patterns:
+          - "*"
```
boholder, line 26: Will this part work without `patterns`? Seems like equal to "include all"

gradle/libs.versions.toml
```diff
@@ -0,0 +1,39 @@
...
+junitJupiter = { module = "org.junit.jupiter:junit-jupiter", version.ref = "junitJupiter" }
+junitJupiterParams = { module = "org.junit.jupiter:junit-jupiter-params", version.ref = "junitJupiter" }
+log4jApi = { module = "org.apache.logging.log4j:log4j-api", version.ref = "log4j" }
+log4jCore = { module = "org.apache.logging.log4j:log4j-core", version.ref = "log4j" }
+lombok = "org.projectlombok:lombok:1.18.30"
+mockitoCore = { module = "org.mockito:mockito-core", version.ref = "mockito" }
+mockitoJunitJupiter = { module = "org.mockito:mockito-junit-jupiter", version.ref = "mockito" }
+
+[plugins]
+architecturyLoom = { id = "dev.architectury.loom", version = "1.9-SNAPSHOT" }
```
boholder, line 37: Hope you add the comment back:
```suggestion
# From developers of architectury: "shouldn't hardcode the loom version, just put *-SNAPSHOT"
# ref: https://maven.architectury.dev/dev/architectury/architectury-loom/
architecturyLoom = { id = "dev.architectury.loom", version = "1.9-SNAPSHOT" }
```

build.gradle
boholder: By "this block" you mean `repositories`? But there are still contents under this block, and partly duplicate with `settings.gradle`. Is there a way to keep only one copy of repo content?

common/build.gradle
```diff
@@ -1,67 +1,39 @@
...
-    testAnnotationProcessor "org.openjdk.jmh:jmh-generator-annprocess:1.37"
+    testImplementation libs.jmhCore
+    testAnnotationProcessor libs.jmhGeneratorAnnprocess
 }
 
 test {
     useJUnitPlatform()
-    // Since JDK 21
-    // ref: https://stackoverflow.com/questions/77951485/getting-a-java-agent-has-been-loaded-warning-in-intellij-after-upgrading-jdk-17
     jvmArgs("-XX:+EnableDynamicAgentLoading")
```
boholder, line 36: Add explanation back, for the future.
```suggestion
    // Suppress JVM warning about mockito's behavior when running unit tests with IntelliJ
    // ref: https://stackoverflow.com/questions/77951485/getting-a-java-agent-has-been-loaded-warning-in-intellij-after-upgrading-jdk-17
    jvmArgs("-XX:+EnableDynamicAgentLoading")
```

fabric/build.gradle
boholder: > Do we even directly depend on either of these in the `fabric` subproject?

No, [the fabric version of archi api depends on the fabric api](https://github.com/architectury/architectury-api/blob/cd74016d85b562affb21890cb8796b1e34ac2ca9/fabric/build.gradle#L28-L29), and [the only class](https://github.com/minecraft-access/minecraft-access/blob/dev/fabric/src/main/java/org/mcaccess/minecraftaccess/fabric/MainClassFabric.java) under `fabric` doesn't directly use fabric api, so yes I think `include` is enough for fabric loader and fabric api. But I think leave archi api for future isn't bad.

gradle/libs.versions.toml
boholder, line 1: maybe you can add more spaces under `[library]`, and the position of `jade` and `log4j` aren't quite accurate.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,39 @@
...
+architecturyPlugin = { id = "architectury-plugin", version = "3.4-SNAPSHOT" }
+gitVersion = { id = "com.palantir.git-version", version = "3.1.0" }
+shadow = { id = "com.gradleup.shadow", version = "8.3.5" }
```
boholder: How about add shadow comment to here and remove comments in fabric and neoforge build file
```suggestion
# The shadow gradle plugin assists with repackaging a 3rd-party library jar within your own jar
# In addition to ensuring that your end-users have the library available when they use your own thing,
# it also helps avoid collisions with other things that are also using the same library.
# ref: https://gist.github.com/SizableShrimp/949e7c219bfc94487a45226b64ac7749
shadow = { id = "com.gradleup.shadow", version = "8.3.5" }
```

boholder (changes requested):
I haven't run `gradle build` to see the result, I'll run it later.

settings.gradle
BrailleBennett: What order do you think it should be in? It's almost just alphabetical.

build.gradle
BrailleBennett: As far as I know, there is no way to remove the coppied code. Dependabot checks settings.gradle for repositories, and when gradle runs it won't use settings.gradle because Architectury loom provides a list of repositories.

common/build.gradle
BrailleBennett: I disagree, people don't care at what point a change was made. It doesn't provide any useful info about your implementation, and I'm not sure what purpose it serves.

boholder: Then delete this `jvmArgs("-XX:+EnableDynamicAgentLoading")` config

boholder: And run the unit test

boholder: Ok, modified

fabric/build.gradle
BrailleBennett: > This `.toString()` is quite messy. Do we even directly depend on either of these in the `fabric` subproject? If not we should be able to completely remove the `modImplementation` and just use `include`.

Removing the mod implementation keyword causes the mod to not run in fabric. This is out of scope for this PR anyways, as these were previously declared as mod implementations and this PR does not change that fact.

neoforge/build.gradle
BrailleBennett: > This `.toString()` is quite messy. Do we even directly depend on either of these in the `neoforge` subproject? If not we should be able to completely remove the `modImplementation` and just use `include`.

Removing the mod implementation keyword causes the mod to not run in NeoForge. This is out of scope for this PR anyways, as these were previously declared as mod implementations and this PR does not change that fact.

gradle/libs.versions.toml
BrailleBennett: As we don't describe the usage of any other plugin or even library for that matter, I removed the comment entirely. As we are using the plugin like normal, a user can look up what shadow is on their own.

gradle/libs.versions.toml
BrailleBennett: What do you mean by they're not accurate?

fabric/build.gradle
boholder: > Removing the mod implementation keyword causes the mod to not run in fabric.

You mean change `modImplementation` to `include`, or directly delete these lines?

gradle/libs.versions.toml
boholder: > As we don't describe the usage of any other plugin or even library for that matter, I removed the comment entirely. As we are using the plugin like normal, a user can look up what shadow is on their own.

This is for archi loom, not shadow. And, it's very hard to directly search out this information, that's why I left comments just for my own in future.

gradle/libs.versions.toml
boholder: jade is a third party mod that we support, log4h is not unit test dependency

boholder:
What's the purpose of commit changes about #468

BrailleBennett:
> What's the purpose of commit changes about #468 

There would have just been conflicts on conflicts when this got merged, and it was only like a few lines over all that I removed, so I thought it was appropriate to put it in this PR for the reason that it is related. I also split the contributing.md changes off into #474 so we can spend longer on them and make sure that we're happy with it before merging.

gradle/libs.versions.toml
```diff
@@ -20,7 +21,7 @@ jadeNeoForge = "maven.modrinth:jade:17.3.0+neoforge"
 minecraft = "com.mojang:minecraft:1.21.4"
 neoForge = "net.neoforged:neoforge:21.4.89-beta"
 parchment = "org.parchmentmc.data:parchment-1.21.4:2025.02.16"
-# Testing
+# Secondary Dependencies
 assertjCore = "org.assertj:assertj-core:3.24.2"
```
boholder: No, assertj is for testing, sorry

gradle/libs.versions.toml
BrailleBennett: Sorry, this is what I meant to reply to the other day. Since we use shadow as normal, people can look up what it is.

gradle/libs.versions.toml
BrailleBennett: My bad, I replied to the wrong conversation the other day. I think that we should probably pin all dependencies moving forward to keep consistent build output. Since it is automatic with dependabot, we will get the latest versions weekly, which should be more than enough for this size of a project.

boholder (dismissed):

settings.gradle
sgmedev: The main things I find weird is that `mavenCentral()` and the `exclusiveContent` are in the middle. AFAIK the order of `exclusiveContent` doesn't matter so it would make more sense to put it at the very end instead of smack bang in the middle of everything else. And as for `mavenCentral()` we should make up our mind as to whether we want it to be tried before or after other repos, being before some but after others doesn't make much sense.

neoforge/build.gradle
sgmedev: This is very much within scope. The main problem isn't the `modImplementation`, that was mearly a suggested fix. The problem is that converting stuff to and from strings like this is a major antipattern. Have you tried splitting it into two separate statements as I suggested to you a while back?

boholder:
Sorry but, one of my questions remain unanswered: https://github.com/minecraft-access/minecraft-access/pull/471#discussion_r1967294180

.github/dependabot.yml
BrailleBennett: > Will this part work without `patterns`? Seems like equal to "include all"

As far as I can tell, it is required when you're using grouped dependencies. Feel free to prove me wrong though.

boholder: > it is required when you're using grouped dependencies.

But the upper one, `gh-actions` doesn't have `patterns` config. I can't find a linter for dependabot config to prove which is true, sorry.

doc: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--

boholder (changes requested):

boholder (dismissed):

boholder (approved):
Nice, really, the comments you write will benefit readers in future.

sgmedev (approved):

PR #473, opened 2025-02-25
refactor: move item speaking injection point for neoforge out again

boholder:
## Why picky on that warning?

In development we debug run the game a lot, at every running this warning will set the focus of IDE debug panel to itself and I need to manually focus on debug output, it's annoying.

## Any other way to suppress the warning?

I tried things like `@SuppressWarnings("UnresolvedMixinReference")` but none of them works. All my want is suppress this exact warning, so easier ways are welcomed.

## Any other way to reuse the mixin code?

Sadly, nope. The target class, `Gui` doesn't have parents so I can't use [Mixin Inheritance](https://wiki.fabricmc.net/tutorial:mixinheritance).

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/mixin/GuiMixin.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/mixin/GuiMixin.java
@@ -25,7 +25,7 @@ import java.util.List;
  * Narrates titles
  */
 @Mixin(Gui.class)
-public class GuiMixin {
+abstract class GuiMixin {
     @Shadow
     private int toolHighlightTimer;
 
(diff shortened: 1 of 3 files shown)


BrailleBennett (changes requested):
I feel like this is a pretty silly reason to add platform specific mixins back and sets a horrible precedence for the project moving forward. The reason we switch to mojang mappings was to completely remove platform specific code, and the only reason that we're adding it back is because of a quirk in an IDE that isn't related to the player experience in any way.

BrailleBennett:
If we can't handle this with a mixin, then use an event handler from Architectury API.

boholder:
> If we can't handle this with a mixin, then use an event handler from Architectury API.

~~Can you provide some guides about how to achieve this?~~

https://docs.architectury.dev/api/events
No such "held item changes" event
https://github.com/architectury/architectury-api/blob/cd74016d85b562affb21890cb8796b1e34ac2ca9/common/src/main/java/dev/architectury/event/events/common/PlayerEvent.java

sgmedev (changes requested):
We should be avoiding duplicate and platform specific code. Please find some other way to fix this error, either by finding a common injection point, avoiding mixins alltogether and using an event handler, or otherwise.

boholder:
> We should be avoiding duplicate and platform specific code. Please find some other way to fix this error, either by finding a common injection point, avoiding mixins alltogether and using an event handler, or otherwise.

How do you resolve this problem?

PR #477, opened 2025-03-01
Fixed Mixin warning

sgmedev:
Fixes the Mixin warning by moving the offending code to the tick event handler.

There was nothing special about the offending injection point, all that matters is that we call the `SpeakHeldItem#speakHeldItem()` method every tick.


boholder (approved):
Thank you for resolving this problem

BrailleBennett (approved):

PR #482, opened 2025-03-09
Added scroll keys to chat screen

BrailleBennett:
# Changelog

## New Features
- You can now scroll through old chat messages. View the [keybinding docs](https://docs.mcaccess.org/keybindings/#speak-chat-messages) for the commands
- You can now use `alt` + `0` or `alt` + `numpad 0` to hear the 10th message on a chat page

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/mixin/ChatScreenMixin.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/mixin/ChatScreenMixin.java
@@ -28,14 +28,14 @@ public class ChatScreenMixin {
     private static final Component USAGE_TEXT = Component.translatable("chat_screen.usage");
 
     @Unique
-    private static int minecraft_access$chatMessagePage;
+    private static int minecraft_access$currentChatMessagePage;
 
     @Shadow
     protected EditBox input;
 
     @Inject(at = @At("HEAD"), method = "init")
     private void init(CallbackInfo ci) {
-        minecraft_access$chatMessagePage = 0;
+        minecraft_access$currentChatMessagePage = 0;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


BrailleBennett:
The check that failed is because I force pushed the compile error, since it was literally just one extra space in the init method signature, but the workflow doesn't pull when you force push, so it says the build failed. The test checks passed though.

common/src/main/resources/assets/minecraft_access/lang/en_us.json
```diff
@@ -59,6 +59,7 @@
   "minecraft_access.gui.access_menu_config_menu.button.fluid_detector_button": "Fluid Detector Configs",
   "minecraft_access.gui.camera_controls_config_menu.button.modified_rotating_angle": "Modified Rotation Angle",
   "minecraft_access.gui.camera_controls_config_menu.button.normal_rotating_angle": "Normal Rotation Angle",
+  "minecraft_access.gui.chat_screen.showing_message_range": "Showing messages %d threw %d",
```
sgmedev: Typo, can be fixed later via Weblate though.
```suggestion
  "minecraft_access.gui.chat_screen.showing_message_range": "Showing messages %d through %d",
```

sgmedev (dismissed):

sgmedev (dismissed):

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ChatScreenMixin.java
```diff
@@ -69,13 +78,48 @@ private void keyPressed(int keyCode, int scanCode, int modifiers, CallbackInfoRe
...
+        int numMessages = ((ChatComponentAccessor) Minecraft.getInstance().gui.getChat()).getAllMessages().size();
+        int oldOffset = minecraft_access$chatMessageOffset;
         if (Screen.hasAltDown()) {
+            if (InputConstants.isKeyDown(window, InputConstants.KEY_GRAVE) || InputConstants.isKeyDown(window, GLFW.GLFW_KEY_KP_MULTIPLY)) {
+                if (Screen.hasControlDown()) {
+                    minecraft_access$chatMessageOffset = Math.max(0, ((numMessages - 1) / 10) * 10);
+                } else {
+                    minecraft_access$chatMessageOffset = 0;
+                }
+            } else if ((InputConstants.isKeyDown(window, InputConstants.KEY_EQUALS) || InputConstants.isKeyDown(window, GLFW.GLFW_KEY_KP_ADD)) && minecraft_access$chatMessageOffset >= 10) {
```
boholder: I'm sorry but I'm confused by the if conditions and logic, and I can't give improvement suggestions since I haven't fully understood yet. Like this `&& minecraft_access$chatMessageOffset >= 10`, I don't know. Hope you can extract more reusable helper methods and boolean variables to make the code describes itself.

BrailleBennett: I feel like that one is literally one of the most obvious conditions. You obviously don't want to subtract 10 if the chat message offset isn't ten or greater. If that's not checked, you'll end up in negative values.

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ChatScreenMixin.java
```diff
@@ -69,13 +78,48 @@ private void keyPressed(int keyCode, int scanCode, int modifiers, CallbackInfoRe
...
+                minecraft_access$chatMessageOffset -= 10;
```
boholder: So... Does using `Math.max()` here to get at most zero result also fulfill the purpose? And other same places through the code. Wait, I know a better method, [clamp](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Math.html#clamp(long,int,int))

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ChatScreenMixin.java
```diff
@@ -69,13 +78,43 @@ private void keyPressed(int keyCode, int scanCode, int modifiers, CallbackInfoRe
      */
     @Unique
     private static boolean minecraft_access$repeatPreviousChatMessage(int keyCode) {
+        long window = Minecraft.getInstance().getWindow().getWindow();
+        int numMessages = ((ChatComponentAccessor) Minecraft.getInstance().gui.getChat()).getAllMessages().size();
+        int oldChatMessagePage = minecraft_access$chatMessagePage;
```
boholder: Much better after using `clamp`, thanks, now I can understand the code.
There is one question here, I wonder if the code would be more clear if we use a new local variable for storing the page changes, instead of caching old value then use the class field `minecraft_access$chatMessagePage`, and comparing the new value with saved old value.
BTW, `mca$chatMessagePage` is a shorter but valid `@Unique` name.

boholder (changes requested):
1. The changelog isn't quite clear to me, I can't understand the usage until seeing the code... maybe you can simply mention the new feature with one sentence, following with a link to the keybindings doc.
2. Don't forget to change the doc.

Nothing left, the new code suggestion is optional.

boholder (dismissed):

boholder (approved):

PR #491, opened 2025-03-22
Speak version on init

Molitvan:
# Changelog

## New Features
- The mod will now say its version when starting up


boholder (approved):
It's weird that the change doesn't work on my dev environment, but works on production environment

PR #494, opened 2025-03-23
fix: cloth config narration

boholder:
.


BrailleBennett (approved):
Cloth gradle will be fixed in following PR.

PR #500, opened 2025-03-26
Fix: update other languages I18N file to match the en_us content

boholder:
# Changelog

## Translation Changes

- Existing configuration translation has been applied to new cloth config screen
- Chinese translation has been updated to match the latest content


boholder:
Tested in dev env, game doesn't crush, checked the English and Chinese translation of cloth config.

BrailleBennett (changes requested):
Thanks for working on this! I see what you were going for with cleaning up the Weblate alerts, but I think this might cause more trouble than it solves. Since it touches most of the translation files, it’s likely to lead to a lot of merge conflicts. The alerts in Weblate are basically already a form of a merge conflict since Weblate works off its own git branch. I think the only reason it hasn’t cleaned things up yet is that we last pulled from Weblate on Christmas Day. Once we sync things up again, it should handle most of these issues.
The only thing that might be something we need to fix manually is the cloth config translations in other languages since the key names changed. That can probably be done through weblate though.
Definitely appreciate the effort, though! Let me know if you’d like to chat more about it.

boholder:
@BrailleBennett Wow, thanks for letting me know that weblate has its own git repo, it's the first time I know this. My opinion is, whatever current weblate repo is, apply this pr's result to it, this pr will resolve all current conflicts and gives us a clean baseline. Renaming keys with regex and `sed` is far more fast than on weblate, plus this pr has resolved key renaming problem which introduced by switching to cloth config.
I wonder if there is possible to dry-run applying this pr to weblate repo and see the result.

sgmedev:
This is in direct contravention to how we work with Weblate and makes the merge conflicts four times worse. Firstly, you cannot fix Git merge conflicts by further modifying files in this way, you have to actually merge to fix the merge conflicts. Also, the only files currently with merge conflicts is `en_us.json`, when I attempt to merge Weblate into this branch all the language files other than `vi_vn.json` conflict.

Weblate already has changes to a similar effect of 413409877fd20eb00ee3eb1696a847ed0334fb39, 54629ff997d74e04999bc5b64025c7e31044c03f, and 5e8bfa8c684834e08b664b84744b66a0d9432f5c. The reason the changes aren't in the repository yet is because we haven't had a release since Christmas Day last year.

de5e89d8788eaa3fe93baae114d48244b7c38c0b, e915796eebac489ecdd261381fac29f0993f9f92, and 35412787516cddeaaea3c7aa5a338efeb8d7e140 should've been done in Weblate if they haven't already been done.

I might look into cherry-picking and squashing bd544fea6b3ed91f1b0cb2bb16563feb64a51147, d8fc5f0c771827c0111bb374e9a7817774fc52ee, and cc95401bfa1d1f804fc40640481fbe11ee2fd32c because they rename keys which is OK, though they probably should've been included in their respective pull requests. But I will have to close this because it's easier to cherry-pick those few good commits than to deal with yet more merge conflicts.

It may also be worth noting that the underlything repository that Weblate uses is read-only. The only way to change anything in it is either by pushing to our `dev` branch which it will then pull and rebase onto or to make changes through the UI.

boholder:
@TheSuperGamer20578 glad to hear that this pr helps. Totally understand the close action after reading your decent explaining, thanks for explaining it to me.

PR #505, opened 2025-03-29
Fix/misc code

BrailleBennett:
# Changelog
## Bug Fixes
- The mod will now narrate the new selected item in your hotbar, even if the previous slot you were on matches
- The mod will no longer apply mouse simulation while typing or while in the keybinds controls menu
- The log4j.xml dev configuration file will no longer be bundled in the production mod jar


log4j.xml
```diff
@@ -0,0 +1,38 @@
...
+    <Appenders/>
+    <Loggers>
+        <Root level="OFF">
+            <AppenderRef ref="SysOut" level="ALL"/>
+            <AppenderRef ref="DebugFile" level="ALL"/>
+            <AppenderRef ref="LatestFile" level="INFO"/>
+            <AppenderRef ref="ServerGuiConsole" level="INFO"/>
+        </Root>
+        <Logger name="org.mcaccess.minecraftaccess" level="DEBUG"/>
+        <Logger name="org.mcaccess.minecraftaccess.compat" level="DEBUG"/>
```
boholder, line 16: Listing all the features here isn't a good idea, because it adds another place to remember to change while changing the code. What's the purpose behind these lines? @TheSuperGamer20578

sgmedev: That is so that it's easy for people to quickly change the log level of individual features when they are developing. If for example the fall detector is spamming they can simply find the fall detector line and change the level without needing to understand fully how log4j works. If its not always kept up to date nothing bad happens at all, I have a catch all setting everything from mcaccess to `DEBUG` so if a new feature isn't added here it will just default to `DEBUG` and if a feature is removed log4j should simply ignore it. We also don't get entirely new features too often so I don't think it's going to be a problem.

boholder: > That is so that it's easy for people to quickly change the log level of individual features when they are developing. If for example the fall detector is spamming they can simply find the fall detector line and change the level without needing to understand fully how log4j works.

Yes, I've guessed about your purpose might be this, but I'd better waiting for your response to avoiding getting things complex. So... how about adding comments like "you can suppress/change log level of features by <example>" and remove those feature log level lines?

BrailleBennett: We believe the feature is self explanatory. You just change the word debug to the word info or the word trace. If you don't know how it works, don't touch it and it will keep working the exact same way it works right now.

sgmedev (approved):

PR #509, opened 2025-03-30
chore: suppress @Overwrite warning raised by loom when build, 

boholder:
by adding javadoc


BrailleBennett (approved):

BrailleBennett:
It's not the most ideal, but it does make IntelliJ happy, and it doesn't hurt to have it.

PR #513, opened 2025-03-30
Feature/gamemode player status

### LINKED ISSUE #485: Make player status and warning game mode contextual
In creative, the player status key should simply speak the game mode name, this can be taken from the vanilla I18N text. We should also suppress all warning logic in creative as well, as the health, hunger, etc. are all not shown on the screen for sited people. In addition in adventure mode we should append the player status with an announcement that the player is in adventure mode, as sighted players can tell if they are or not based on their game block curser. In spectator mode, most of the mods functionality should be disabled, as it hardly works correctly in that game mode.

BrailleBennett:
Closes #485
# Changelog

## New Features
- Player Status now reports the game mode when pressed

## Feature Updates

- Player Status is now no longer available in Creative mode or Spectator mode to match sighted player's view
- Player Warnings no longer trigger in Creative mode or Spectator mode to match sighted player's view
- Some functions will no longer attempt to run if they are not relevant to the current game mode

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
@@ -127,7 +127,7 @@ public class MainClass {
             }
         }
 
-        if (playerWarnings != null && config.playerWarnings.enabled)
+        if (playerWarnings != null && config.playerWarnings.enabled && (PlayerUtils.isSurvival()|| PlayerUtils.isAdventure()))
             playerWarnings.update();
 
         if (accessMenu != null && config.accessMenu.enabled)
--- a/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
@@ -139,8 +139,7 @@ public class MainClass {
         // POI Marking will handle POI Scan and POI Locking features inside it
         POIMarking.getInstance().update();
 
-        if (!PlayerUtils.isSpectator())
-            FallDetector.getInstance().update();
+        FallDetector.getInstance().update();
 
         HUDStatus.getInstance().update();
 
--- a/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
@@ -127,7 +127,7 @@ public class MainClass {
             }
         }
 
-        if (playerWarnings != null && config.playerWarnings.enabled && (PlayerUtils.isSurvival()|| PlayerUtils.isAdventure()))
+        if (playerWarnings != null && config.playerWarnings.enabled && (PlayerUtils.isSurvival() || PlayerUtils.isAdventure()))
             playerWarnings.update();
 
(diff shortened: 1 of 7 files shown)


common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
```diff
@@ -127,7 +127,7 @@ public static void clientTickEventsMethod(Minecraft minecraftClient) {
             }
         }
 
-        if (playerWarnings != null && config.playerWarnings.enabled)
+        if (playerWarnings != null && config.playerWarnings.enabled && (PlayerUtils.isSurvival()|| PlayerUtils.isAdventure()))
```
sgmedev:
```suggestion
        if (playerWarnings != null && config.playerWarnings.enabled && (PlayerUtils.isSurvival() || PlayerUtils.isAdventure()))
```

common/src/main/java/org/mcaccess/minecraftaccess/utils/PlayerUtils.java
```diff
@@ -243,4 +245,39 @@ public static boolean isPlayerTyping() {
...
+        if (client == null || client.gameMode == null) return false;
+        GameType currentGameMode = client.gameMode.getPlayerMode();
+        return currentGameMode == GameType.SURVIVAL;
+    }
+
+    public static boolean isHardCore() {
+        Minecraft client = Minecraft.getInstance();
+        if (client == null || client.level == null) return false;
+        LevelData levelData = client.level.getLevelData();
+        return levelData != null && levelData.isHardcore();
```
boholder, line 281: Since original code doesn't have hardcore (it's also survival I guess), you can create an inner enum class in player util for game mode, and add methods like:
- `GameMode getGameMode()`, return the enum, you can use it for switch matching like what you did in `addGameMode` method.
-  `boolean hasStateHud()` for check whether to narrate suvival related thing like player status.

The enum approach encapsulate the mode check logic into enum class, simplify the logic in features.
I haven't check the code of `GameType`, I wonder if it has such methods

sgmedev (changes requested):
Just to confirm, is it intentional that the gamemode is read when using alt+r in survival? It just sounds a bit odd imo having something that usually won't change during normal gameplay in the conditional status message.

boholder:
> It just sounds a bit odd imo having something that usually won't change during normal gameplay in the conditional status message.

@BrailleBennett How about make this game mode narration part being triggered when game mode switching (F3 + F4)? It's similar to HUD narration, just notify player what they're doing.

BrailleBennett:
> Just to confirm, is it intentional that the gamemode is read when using alt+r in survival? It just sounds a bit odd imo having something that usually won't change during normal gameplay in the conditional status message.

It is, only because it's at the end of your message. But I see your point. I can remove it from there.

BrailleBennett:
> 
> 
> 
> @BrailleBennett How about make this game mode narration part being triggered when game mode switching (F3 + F4)? It's similar to HUD narration, just notify player what they're doing.

That already speaks the switched game mode.

sgmedev (approved):

Issue #302, opened 2024-07-25
[Linux] Replace `xdotool` with an alternate to better support wayland


khanshoaib3:
[ydotool](https://github.com/ReimuNotMoe/ydotool) is an alternative to xdotool which not only supports wayland but also x11 and others. A much better choice than xdotool.

boholder:
This info might be helpful to this issue, posted by @Aya-Komichi in https://github.com/khanshoaib3/minecraft-access/issues/310#issuecomment-2304359402_

Aya-Komichi:
In addition to what I mentioned in [#310 (comment)](https://github.com/khanshoaib3/minecraft-access/issues/310#issuecomment-2304359402), I will add here the information I gathered about ydotool.

The tool itself is rather finicky imo, at least on my arch linux based OS installing it is not enough to get things running. This tool relies on a daemon that ideally should run in the background since it doesn't just send things directly to x11 or wayland. In my experience the installation of ydotool comes with a user level systemd service file to automatically start the ydotoold daemon.

From my research, different linux distributions may bring in different service files for this daemon or none at all. The file I got in arch is actually nearly useless since not only it does not start the command with the necessary flags, but at least for the mouse movement to work correctly with ydotool, the service started needs to be a system level service and not a user level one, because the daemon works better with root permissions. Additionally, the socket path and socket ownership flags need to be set appropriately. 

I had to create a system level service file based on the user level service file I was provided. After the necessary edits, the file looks like this:
```
[Unit]
Description=Starts ydotoold service

[Service]
Type=simple
Restart=always
ExecStart=/usr/bin/ydotoold -p <home-path>/.ydotool_socket -o 1000:1000
ExecReload=/usr/bin/kill -HUP $MAINPID
KillMode=process
TimeoutSec=180

[Install]
WantedBy=default.target
```

The <home-path> needs to be replaced by the path to the home directory of the user. The numbers next to the -o flag are the user id and the user group id which respectively, both are 1000 by default. The home path and ids need to be hardcoded since otherwise trying to use the systemd unit specifiers on a system level service will result in the root home directory and ids being used on the command instead. This is probably the reason why the user service included in the package installation isn't that useful by default.

Alternatively, a user can also just run the following command whenever they login:
```sudo -b ydotoold -p "$HOME/.ydotool_socket" -o "$(id -u):$(id -g)"```

[...]

PR #515, opened 2025-04-01
Simulate mouse with java code instead of native tools

### LINKED ISSUE #374: Mouse Simulation: Cursor moves down when clicking
By @emassey0135:

It seems that when you press left bracket to click on an item to pick it up, the y coordinate of the mouse down and mouse up events is different than the y coordinate of the mouse move event right before it, both in my log and in the log sent in mc-help. From looking at the code, it seems like the mouse should be clicked exactly at where its currently at when the user presses the left mouse simulation key. The x coordinate is the same, but the y coordinate is off by about one or two hundred. I can see if its also like that on Windows and Linux.

### LINKED ISSUE #386: Mouse Simulation: Get rid of native mouse invocations
Related closed issue: #39 #22, if this issue can be finished, the #374 can also be resolved.

- [ ] find a way to modify the coordinates that game thinks mouse is.
- [ ] make an API for retrieving in-bound coordinates of given GUI component.
- [x] replace `MouseUtil` with new system.
- [ ] update doc about native mouse API

PR 1: make mouse mock works, replace `Mouse Simulation` feature with the new way.
PR 2: make GUI monk works but not replace whole `MouseUtil`, just replace one use case for each util method.
PR 3: replace the rest and update doc

This mod has achieved mouse simulation: https://github.com/isXander/Controlify

boholder:
Rewrite `MouseUtils` with invoking `MouseHandler`, the original logic for handling mouse events gotten from native API.
Close #386, #374, #302

# Changelog

## Bug Fixes

- Fix cursor moves down when clicking issue on macOS [#374](https://github.com/minecraft-access/minecraft-access/issues/374)

## Others

- Now the mod no longer needs third-party tools to perform mouse operations.


emassey0135:
On MacOS, the left, middle, and right buttons work for breaking, picking, and using blocks, and the scroll buttons work for moving through the hotbar. Pressing one of the scroll buttons often moves more than one slot though. However, in the inventory, I am experiencing the same issue others have experienced, where I cannot pick up an item, even though I hear the sound the Sounds mod makes for this. Turning on the Mac mouse fix option and enabling fullscreen did not fix it, just like someone else experienced with the old code.

emassey0135:
Also, is it possible to move the mouse by calling Minecraft methods, and get rid of all native mouse code? That would make the code a lot simpler, since now it uses both approaches, and would remove the dependency on xdotool on Linux and possibly fix the inventory issues on MacOS.

emassey0135:
On MacOS, when exiting a world, the Multiplayer button is automatically clicked for some reason, every time I tried, whether fullscreen or Mac mouse fix are on or not.

boholder:
> Also, is it possible to move the mouse by calling Minecraft methods, and get rid of all native mouse code?

Affirmative, that's the next goal, although that's tough.

boholder:
> On MacOS, when exiting a world, the Multiplayer button is automatically clicked for some reason, every time I tried, whether fullscreen or Mac mouse fix are on or not.

Does this also happens in old code?

emassey0135:
> > On MacOS, when exiting a world, the Multiplayer button is automatically clicked for some reason, every time I tried, whether fullscreen or Mac mouse fix are on or not.
> 
> Does this also happens in old code?

It did not when I last used it, although when I started Minecraft, focus would often move to a different window and sometimes a menu would be opened, like the mouse was being clicked somewhere on screen. So I think something odd is happening with the click triggered on the initial game screen. I will test the old code again though, maybe something changed in 1.21.5 or MacOS 15.4.

boholder:
@BrailleBennett no, this pr's scope doesn't includes gui coordination (mouse move), so it won't fix that issue you linked

BrailleBennett:
> @BrailleBennett no, this pr's scope doesn't includes gui coordination (mouse move), so it won't fix that issue you linked 

The issue occurres when you clicked the mouse. That's the one thing this PR does change. I would assume it would be fixed now.

emassey0135:
> > @BrailleBennett no, this pr's scope doesn't includes gui coordination (mouse move), so it won't fix that issue you linked
> 
> The issue occurres when you clicked the mouse. That's the one thing this PR does change. I would assume it would be fixed now.

This definitely does not fix the issue with inventory controls on MacOS. Perhaps the issue is caused by the mouse not being moved to the exact position it should be, because the log records where the mod attempts to move the mouse but not where it actually goes.

emassey0135:
I created a pull request against your branch, replacing the rest of the native code with a function that moves the mouse with LWJGL and the Minecraft MouseHandler class. I still need to test it on Windows and MacOS, but it works in the inventory and screens on Linux.

emassey0135:
The issue with the inventory is fixed for me with the new changes, although now when you exit a world, the Realms button is automatically pressed for some reason. We should make sure the inventory issue is fixed for others too though because at first it wasn't happening for me.

emassey0135:
The Realms issue also occurs on Windows, and for some reason on both operating systems, when I go to Options, and then Music and Sounds, the Ambient and Environment slider is set to 34% without me doing anything, and pressing tab doesn't navigate through the options right. So this seems to break something related to screens or maybe the menu fix.

BrailleBennett:
> The issue with the inventory is fixed for me with the new changes, although now when you exit a world, the Realms button is automatically pressed for some reason. We should make sure the inventory issue is fixed for others too though because at first it wasn't happening for me.
> 

Are you using the space key by chance?

emassey0135:
> > The issue with the inventory is fixed for me with the new changes, although now when you exit a world, the Realms button is automatically pressed for some reason. We should make sure the inventory issue is fixed for others too though because at first it wasn't happening for me.
> 
> Are you using the space key by chance?

I might have been, but it happens if I press enter too, and whenever I press escape on the Realms screen, or press space or enter on the back button, it almost instantly goes back to the Realms screen.

boholder:
> > > The issue with the inventory is fixed for me with the new changes, although now when you exit a world, the Realms button is automatically pressed for some reason. We should make sure the inventory issue is fixed for others too though because at first it wasn't happening for me.
> > 
> > 
> > Are you using the space key by chance?
> 
> I might have been, but it happens if I press enter too, and whenever I press escape on the Realms screen, or press space or enter on the back button, it almost instantly goes back to the Realms screen.

So, conclusion, when exiting back to main screen, expect nothing happens, but enter the realm screen?
@emassey0135 Can you shortly describe what keys you pressed for exiting, IIRC I pressed exit key with real mouse last night, nothing happens. I'd try to reproduce this on Windows once I know the operations.

emassey0135:
> > > > The issue with the inventory is fixed for me with the new changes, although now when you exit a world, the Realms button is automatically pressed for some reason. We should make sure the inventory issue is fixed for others too though because at first it wasn't happening for me.
> > > 
> > > 
> > > Are you using the space key by chance?
> > 
> > 
> > I might have been, but it happens if I press enter too, and whenever I press escape on the Realms screen, or press space or enter on the back button, it almost instantly goes back to the Realms screen.
> 
> So, conclusion, when exiting back to main screen, expect nothing happens, but enter the realm screen? @emassey0135 Can you shortly describe what keys you pressed for exiting, IIRC I pressed exit key with real mouse last night, nothing happens. I'd try to reproduce this on Windows once I know the operations.

I tab to single or multi-player and press space or enter, then press escape to go back to the title screen, or press Shift+tab to the back button and press space or enter on it. Then the realms screen opens, and pressing escape from there instantly returns me to the Realms screen, after going to the title screen first. At least on Linux, this doesn't happen on fullscreen, only when fullscreen is disabled. The issue with music and sounds options mentioned above also only occurs on Linux. However, the inventory works no matter whether the game is fullscreen or not.

emassey0135:
I just looked at features/MenuFix.java, and its easy to see why its doing this. The moveMouseCursor method moves the mouse to 10 plus the coordinate of the window on the screen for both x and y, but the new mouse move function takes coordinates relative to the window and not the screen.

emassey0135:
I fixed it: https://github.com/boholder/minecraft-access/pull/162

BrailleBennett:
#551

PR #521, opened 2025-04-07
Auto unlock from picked up Item

boholder:


common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
```diff
@@ -121,6 +123,16 @@ private void lookAtLockedTarget() {
         }
     }
 
+    private boolean unlockFromPickedUpItem() {
+        if (NarrationUtils.isDroppedItem(lockedOnEntity)
+            && lockedOnEntity.getName().getContents() instanceof TranslatableContents trans
+            && trans.getKey().equals("block.minecraft.air")) {
```
sgmedev, line 129: This seems a bit convoluted, would it be possible to check directly instead of going through the translation key?

sgmedev (changes requested):

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
boholder: I also think it's complexing, but it is the only way I found to retrieve this information from item entity. You can extend the scan interval to very long, pick up the dropped item, then debug break point at here, see if you can find a simpler way to get the "air" state.

sgmedev:
Superceeded by #569

Issue #369, opened 2024-11-03
Refactor I18N


sgmedev:
Currently we use a mixin to alter the `I18n` class to fall back to English and to use named placeholders. English fall-back is vanilla behaviour and named placeholders should be implemented in a wrapper as to avoid interfering with other mods, therefore the `I18n` mixin should be removed.

Additionally, most strings use `String.format()`-style placeholders which can be unclear to translators and they are used unindexed which makes it impossible for translators to reorder. Named placeholders should be used everywhere.

boholder:
It worth note that, the [NamedFormatter](https://github.com/minecraft-access/minecraft-access/blob/dev/common/src/main/java/org/mcaccess/minecraftaccess/utils/NamedFormatter.java) I made, to format strings like `{foo} at {bar}`, this type of string will be marked by IntelliJ MC dev plugin, says its format is illegal (invalid). That paused my step to replace every i18n text to that format.

sgmedev:
> It worth note that, the [NamedFormatter](https://github.com/minecraft-access/minecraft-access/blob/dev/common/src/main/java/org/mcaccess/minecraftaccess/utils/NamedFormatter.java) I made, to format strings like `{foo} at {bar}`, this type of string will be marked by IntelliJ MC dev plugin, says its format is illegal (invalid). That paused my step to replace every i18n text to that format.

I don't see any warnings in mine, where does it show you that warning?

boholder:
> I don't see any warnings in mine, where does it show you that warning?

See in IDE

https://github.com/minecraft-access/minecraft-access/blob/ac9a286c752aa164bdbd03e1474ff611023f5d21/common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java#L121

https://github.com/minecraft-access/minecraft-access/blob/ac9a286c752aa164bdbd03e1474ff611023f5d21/common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java#L68

sgmedev:
> > I don't see any warnings in mine, where does it show you that warning?
> 
> See in IDE
> 
> [minecraft-access/common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java](https://github.com/minecraft-access/minecraft-access/blob/ac9a286c752aa164bdbd03e1474ff611023f5d21/common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java#L121)
> 
[...]

PR #525, opened 2025-04-12
Fix enchanting table

### LINKED ISSUE #467: Enchanting table UI reports wrong enchantments
When trying to enchant something, Minecraft Access will read out the wrong enchantments, for example Blast Protection on a sword. Using OCR on the same UI reveals that the enchantments are actually normal.

Molitvan:
closes #467 

Other than what I wrote in the changelog, I also refactored the enchanting table code to be more readable aka I gave all the variables normal names.

# Changelog

## Feature Updates
- You will now be able to see when you don't have enough lapis or levels when enchanting
## Bug Fixes
- Fixed enchanting table UI reading wrong enchantments

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/features/inventory_controls/GroupGenerator.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/features/inventory_controls/GroupGenerator.java
@@ -368,16 +368,21 @@ public class GroupGenerator {
                 Optional<Holder.Reference<Enchantment>> enchantment = Minecraft.getInstance().level.registryAccess().lookupOrThrow(Registries.ENCHANTMENT).get(enchantmentId);
                 if (enchantment.isEmpty()) break;
 
-                StringBuilder clueText = new StringBuilder(Component.translatable("container.enchant.clue", Enchantment.getFullname(enchantment.get(), level)).getString());
+                StringBuilder clueText = new StringBuilder(Component.translatable("container.enchant.clue", Enchantment.getFullname(enchantment.get(), level)).getString()); // this breaks when using I18n.get() for some reason so Component.translatable() is being used instead
                 if (!PlayerUtils.isCreative()) {
                     if (Minecraft.getInstance().player.experienceLevel < requiredLevel) {
-                        clueText.append(Component.translatable("container.enchant.level.requirement", requiredLevel).getString());
+                        clueText.append(I18n.get("container.enchant.level.requirement", requiredLevel));
                     } else {
-                        String lapisCostText = cost == 1 ? Component.translatable("container.enchant.lapis.one").getString() : Component.translatable("container.enchant.lapis.many", cost).getString();
-                        if (enchantmentScreenHandler.getGoldCount() < cost) lapisCostText = "()" + I18n.get("minecraft_access.other.missing") + ") " + lapisCostText;
+                        String lapisCostText = cost == 1 ? I18n.get("container.enchant.lapis.one") : I18n.get("container.enchant.lapis.many", cost);
+                        if (enchantmentScreenHandler.getGoldCount() < cost) {
+                            lapisCostText = String.format("(%s) %s", I18n.get("minecraft_access.other.missing"), lapisCostText);
+                        }
                         clueText.append(lapisCostText);
-                        String xpCostText = cost == 1 ? Component.translatable("container.enchant.level.one").getString() : Component.translatable("container.enchant.level.many", cost).getString();
-                        if (PlayerUtils.getExperienceLevel() < cost) xpCostText = "()" + I18n.get("minecraft_access.other.missing") + ") " + xpCostText;
+
+                        String xpCostText = cost == 1 ? I18n.get("container.enchant.level.one") : I18n.get("container.enchant.level.many", cost);
+                        if (PlayerUtils.getExperienceLevel() < cost) {
+                            xpCostText = String.format("(%s) %s", I18n.get("minecraft_access.other.missing"), xpCostText);
+                        }
                         clueText.append(xpCostText);
                     }
                 }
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/mcaccess/minecraftaccess/features/inventory_controls/GroupGenerator.java
```diff
@@ -358,32 +359,30 @@ public static List<SlotsGroup> generateGroupsFromSlots(AbstractContainerScreenAc
...
+                    if (Minecraft.getInstance().player.experienceLevel < requiredLevel) {
+                        clueText.append(Component.translatable("container.enchant.level.requirement", requiredLevel).getString());
                     } else {
-                        MutableComponent mutableText = m == 1 ? Component.translatable("container.enchant.lapis.one") : Component.translatable("container.enchant.lapis.many", m);
-                        clueText.append(mutableText.withStyle(i >= m ? ChatFormatting.GRAY : ChatFormatting.RED).getString());
-                        MutableComponent mutableText2 = m == 1 ? Component.translatable("container.enchant.level.one") : Component.translatable("container.enchant.level.many", m);
-                        clueText.append(mutableText2.withStyle(ChatFormatting.GRAY).getString());
+                        String lapisCostText = cost == 1 ? Component.translatable("container.enchant.lapis.one").getString() : Component.translatable("container.enchant.lapis.many", cost).getString();
```
sgmedev: This should use `I18n.get()`, same below. It looks like it was like this before though so if you want to leave it that's fine too.

common/src/main/java/org/mcaccess/minecraftaccess/features/inventory_controls/GroupGenerator.java
```diff
@@ -358,32 +359,30 @@ public static List<SlotsGroup> generateGroupsFromSlots(AbstractContainerScreenAc
...
+                        if (enchantmentScreenHandler.getGoldCount() < cost) lapisCostText = "()" + I18n.get("minecraft_access.other.missing") + ") " + lapisCostText;
+                        clueText.append(lapisCostText);
+                        String xpCostText = cost == 1 ? Component.translatable("container.enchant.level.one").getString() : Component.translatable("container.enchant.level.many", cost).getString();
+                        if (PlayerUtils.getExperienceLevel() < cost) xpCostText = "()" + I18n.get("minecraft_access.other.missing") + ") " + xpCostText;
```
sgmedev: This line is quite long and complex, it might be better to put the statement within the if statement on its own line. It may also be simpler to use `String.format()` instead of concatenating four things together. The first closing paren also looks like it may be a typo but I could be wrong.

```suggestion
                        if (PlayerUtils.getExperienceLevel() < cost) {
                            xpCostText = String.format("(%s) %s", I18n.get("minecraft_access.other.missing"), xpCostText);
                        }
```

sgmedev (commented):
Overall looks like a massive improvement, thanks for fixing those variable names. I've just spotted a couple minor points of further improvement, though I will still need to open it up properly and test it tomorrow.

Molitvan:
> Overall looks like a massive improvement, thanks for fixing those variable names. I've just spotted a couple minor points of further improvement, though I will still need to open it up properly and test it tomorrow.

There are a couple other places with nonsense variable names for example the villager trading code right below the enchanting table code. Do you think this deserves an issue?

common/src/main/java/org/mcaccess/minecraftaccess/features/inventory_controls/GroupGenerator.java
Molitvan: > This should use `I18n.get()`, same below. It looks like it was like this before though so if you want to leave it that's fine too.

I also thought so, but I assumed there was a reason for it not using I18n.get() already.

common/src/main/java/org/mcaccess/minecraftaccess/features/inventory_controls/GroupGenerator.java
Molitvan: > This line is quite long and complex, it might be better to put the statement within the if statement on its own line. It may also be simpler to use `String.format()` instead of concatenating four things together. The first closing paren also looks like it may be a typo but I could be wrong.

Ugh I forgot about String.format(). Thanks

common/src/main/java/org/mcaccess/minecraftaccess/features/inventory_controls/GroupGenerator.java
```diff
@@ -358,32 +359,35 @@ public static List<SlotsGroup> generateGroupsFromSlots(AbstractContainerScreenAc
...
+            for (int i = 0; i < 3; i++) {
+                int enchantmentId = enchantmentScreenHandler.enchantClue[i];
+                int requiredLevel = enchantmentScreenHandler.costs[i];
+                int level = enchantmentScreenHandler.levelClue[i];
+                int cost = i + 1;
+
+                Optional<Holder.Reference<Enchantment>> enchantment = Minecraft.getInstance().level.registryAccess().lookupOrThrow(Registries.ENCHANTMENT).get(enchantmentId);
+                if (enchantment.isEmpty()) break;
+
+                StringBuilder clueText = new StringBuilder(Component.translatable("container.enchant.clue", Enchantment.getFullname(enchantment.get(), level)).getString()); // this breaks when using I18n.get() for some reason so Component.translatable() is being used instead
```
sgmedev, line 369: The reason that stuff breaks is that `Enchantment.getFullname()` returns a `Component`. This can be fixed by adding `.getString()` to it but that's only marginally better than whats already happening so it doesn't really matter too much, this will be refactored as part of #369 anyway.
```suggestion
                StringBuilder clueText = new StringBuilder(I18n.get("container.enchant.clue", Enchantment.getFullname(enchantment.get(), level).getString()));
```

common/src/main/java/org/mcaccess/minecraftaccess/features/inventory_controls/GroupGenerator.java
sgmedev, line 1: Unused imports:
- `net.minecraft.ChatFormatting`
- `net.minecraft.network.chat.Component`
- `net.minecraft.network.chat.MutableComponent`

sgmedev (changes requested):
Just a few unused imports, otherwise I think this is ready to merge once those are fixed.

sgmedev:
> > Overall looks like a massive improvement, thanks for fixing those variable names. I've just spotted a couple minor points of further improvement, though I will still need to open it up properly and test it tomorrow.
> 
> There are a couple other places with nonsense variable names for example the villager trading code right below the enchanting table code. Do you think this deserves an issue?

If you're able to work on it sometime soon I wouldn't bother, but if you just want to document it for you or someone else to fix later it probably is a good idea to open an issue so that we don't forget.

sgmedev (approved):

PR #534, opened 2025-04-18
Narrator fixes

emassey0135:
This removes the dependency on Flite when running on Linux, and also fixes an infinite loop when running on Pojav Launcher for iOS/Android. On Linux, com.mojang.text2speech.Narrator.getNarrator() tries to load the Flite library and throws an exception if it fails, and the checkStatus function in GameNarrator brings up a dialog asking if you want to continue loading the game. On Pojav Launcher, this dialog causes an infinite loop somehow. Adding a NarratorDummy interface and redirecting getNarrator to it in GameNarratorMixin fixes both of these issues. This PR also enables the narrator by default in OptionsMixin, and disables the custom narration in AccessibilityOnboardingScreen, which tries to load Flite on Linux and is not needed when the narrator is enabled by default.

# Changelog

## New Features

- The narrator is now enabled by default

## Bug Fixes

- Flite is no longer required to be installed on Linux
- The accessibility onboarding screen no longer speaks using the vanilla Minecraft voice instead of Minecraft Access


BrailleBennett (changes requested):
Can you update the docs to remove where it says flight is a required dependency on Linux? I believe that's in the read me and basic setup guide, though a search in all wouldn't be a bad idea.

sgmedev (dismissed):

BrailleBennett (approved):

PR #536, opened 2025-04-20
Only speak if window focused

BrailleBennett:
Narration is now always suppressed if a string is empty and a warning is printed about it.

# Changelog

## Bug Fixes
- Narration of messages other than the initial initializing message are now suppressed when the game is not in focus

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
@@ -171,7 +171,7 @@ public class MainClass {
 
     public static void speakWithNarrator(String text, boolean interrupt) {
         if (Strings.isNotEmpty(text) && Minecraft.getInstance().isWindowActive()) {
-            log.warn("The speaking of string %s with interrupt%s was suppressed".formatted(text, interrupt));
+            log.warn("The speaking of string \"{}\" with interrupt={} was suppressed", text, interrupt);
             return;
         }
         MainClass.interrupt = interrupt;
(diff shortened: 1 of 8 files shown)


common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
```diff
@@ -170,13 +170,11 @@ public static void setScreenReader(ScreenReaderInterface screenReader) {
     }
 
     public static void speakWithNarrator(String text, boolean interrupt) {
+        if (Strings.isNotEmpty(text) && Minecraft.getInstance().isWindowActive()) {
+            log.warn("The speaking of string %s with interrupt%s was suppressed".formatted(text, interrupt));
```
sgmedev: Log calls already have builtin formatting.
```suggestion
            log.warn("The speaking of string \"{}\" with interrupt={} was suppressed", text, interrupt);
```

common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
```diff
@@ -170,13 +170,11 @@ public static void setScreenReader(ScreenReaderInterface screenReader) {
     }
 
     public static void speakWithNarrator(String text, boolean interrupt) {
+        if (Strings.isNotEmpty(text) && Minecraft.getInstance().isWindowActive()) {
```
sgmedev: This condition is backwards.
```suggestion
        if (Strings.isEmpty(text) || !Minecraft.getInstance().isWindowActive()) {
```

sgmedev (changes requested):

sgmedev (approved):

PR #539, opened 2025-04-22
Fix MenuFix

emassey0135:
This fixes the issue with the Realms screen being automatically opened and the issue with Music and Sound options, and probably a lot of issues in menus I haven't found yet.

# Changelog


emassey0135:
Sorry, wrong repository.

PR #541, opened 2025-04-23
Ensure the advancements key is unbound on all platforms

### LINKED ISSUE #376: Keys aren't unbound on neoforge
On neoforge, it appears that the advancements key isn't unbound like it is on fabric on first run of the mod. It is bound to the L key by default.

BrailleBennett:
Fixes #376

# Changelog

## Bug Fixes
- Resolves a bug where the advancements key could remain bound preventing usage of the camera controls


BrailleBennett:
I should maybe add a conditional check to see what the current key is for advancements before I unbind it as well. It could be that a low vision player can read the advancement screen fine.
Thoughts?

sgmedev (approved):

PR #542, opened 2025-04-26
Gradle v8.14

BrailleBennett:
This bumps gradle from 8.13 to 8.14


BrailleBennett:
Compiled jars were tested in production on both fabric and neoforge

sgmedev (approved):

PR #544, opened 2025-04-26
Fix Almost all gradle deprecation warnings

### LINKED ISSUE #472: Resolve gradle deprecation warnings
There are a number of deprecation warnings that we get whenever we run a task on gradle. While some are from plugins we use, most are from our own scripts not being up to standards. This isn't important at the current moment, but eventually gradle 9 will release, and if we resolve these warnings now it is less work that needs to be done then to prepare for the switch along with Loom.

BrailleBennett:
Fixes all of our gradle deprecation warnings. [One warning remains](https://scans.gradle.com/s/irq47kzoi4u3o/deprecations?expanded=WyIzIl0) on both fabric and Neo for `transformProduction`, but the best I can tell, those aren't our tasks and they come from an external plugin like Architectury.
Closes #472

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,7 +73,6 @@ allprojects {
     apply plugin: "java"
     apply plugin: libs.plugins.architecturyPlugin.get().pluginId
     apply plugin: "maven-publish"
-    apply plugin: "base"
 
     base.archivesName = 'minecraft-access'
 
--- a/build.gradle
+++ b/build.gradle
@@ -75,9 +75,7 @@ allprojects {
     apply plugin: "maven-publish"
     apply plugin: "base"
 
-    base {
-        archivesName.set("minecraft-access")
-    }
+    base.archivesName = 'minecraft-access'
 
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -73,8 +73,12 @@ allprojects {
     apply plugin: "java"
     apply plugin: libs.plugins.architecturyPlugin.get().pluginId
     apply plugin: "maven-publish"
+    apply plugin: "base"
```
sgmedev: This is superfluous.

build.gradle
```diff
@@ -73,8 +73,12 @@ allprojects {
...
+
+    base {
+        archivesName.set("minecraft-access")
+    }
```
sgmedev:
```suggestion
    base.archivesName = 'minecraft-access'
```

sgmedev (changes requested):

sgmedev (approved):

PR #547, opened 2025-04-28
Fix and refactor crop narration

### LINKED ISSUE #546: Growth progress of crops is not narrated
In the latest snapshots, crops no longer read their growth status.

Molitvan:
Closes #546 

This bug was introduced in one of the snapshots, so technically no changes from 1.9 to 1.10 thus there is no need for changelog in this case.

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
@@ -32,6 +32,7 @@ import net.minecraft.world.level.block.entity.SignBlockEntity;
 import net.minecraft.world.level.block.entity.SpawnerBlockEntity;
 import net.minecraft.world.level.block.piston.PistonBaseBlock;
 import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.block.state.properties.BlockStateProperties;
 import net.minecraft.world.level.block.state.properties.ComparatorMode;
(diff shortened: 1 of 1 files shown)


BrailleBennett (changes requested):
The stage narration for pitcher plants and nether wart appears to be nonfunctional with this new logic.

common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
```diff
@@ -523,36 +523,16 @@ private static String getSignInfo(SignBlockEntity signEntity, LocalPlayer player
...
-            }
-            case PitcherCropBlock ignored -> {
-                currentAge = blockState.getValue(PitcherCropBlock.AGE);
-                maxAge = PitcherCropBlock.MAX_AGE;
-            }
-            case null, default -> {
-                return new Tuple<>(toSpeak, currentQuery);
-            }
+        if (block instanceof CropBlock) {
+            CropBlock crop = (CropBlock)block;
```
sgmedev: This should use a binding in the instanceof expression instead of a cast. Same below.
```suggestion
        if (block instanceof CropBlock crop) {
```

sgmedev (changes requested):

BrailleBennett:
We should also investigate weather our strings are applicable in all situations. E.G. seedling torch flower crop when it is a single torch flower, but that should be done in weblate later and not in this PR.

sgmedev:
Superceded by #557

PR #551, opened 2025-05-05
Replace native code for mouse simulation with LWJGL and Minecraft calls

### LINKED ISSUE #302: [Linux] Replace `xdotool` with an alternate to better support wayland

### LINKED ISSUE #374: Mouse Simulation: Cursor moves down when clicking

### LINKED ISSUE #386: Mouse Simulation: Get rid of native mouse invocations

emassey0135:

# Changelog

## New Features

- Minecraft Access now works on Wayland on Linux without X11
- Minecraft Access now runs on Pojav Launcher on iOS

## Bug Fixes

- Mouse click positions should no longer be slightly off in MacOS

## Others

- xdotool is no longer required on Linux
- The accessibility permission is no longer required on MacOS


common/src/main/java/org/mcaccess/minecraftaccess/features/MenuFix.java
```diff
@@ -27,16 +29,8 @@
...
-     */
-    @SuppressWarnings("rawtypes")
-    static Class prevScreenClass = TitleScreen.class;
-    /**
-     * The list of screens for which this fix will execute when opened
-     */
-    @SuppressWarnings("rawtypes")
-    private static final List<Class> menuList = new ArrayList<>() {{
+    private static Class<?> prevScreenClass = TitleScreen.class;
+    private static final List<Class<?>> menusNeedFix = new ArrayList<>() {{
```
sgmedev: Would it be better if these were of type `Class<? extends Screen>`?

common/src/main/java/org/mcaccess/minecraftaccess/features/MenuFix.java
```diff
@@ -58,38 +52,30 @@ public class MenuFix {
         add(EditServerScreen.class);
     }};
```
sgmedev: This should use `List.of()`, though since it's immutable an array would be even better.

common/src/main/java/org/mcaccess/minecraftaccess/mixin/MouseHandlerAccessor.java
```diff
@@ -0,0 +1,17 @@
...
+@Mixin(MouseHandler.class)
+public interface MouseHandlerAccessor {
+    @Invoker("onMove")
+    void move(long windowPointer, double x, double y);
+
+    @Invoker("onPress")
+    void press(long windowPointer, int button, int action, int modifiers);
+
+    @Invoker("onScroll")
+    void scroll(long windowPointer, double xOffset, double yOffset);
```
sgmedev: These should probably be renamed to `invokeOnMove`, `invokeOnPress`, and `invokeOnScroll` to avoid inconsistancy and any future confusion, also allows the annotation argument to be removed.

sgmedev (changes requested):

common/src/main/java/org/mcaccess/minecraftaccess/features/MenuFix.java
emassey0135: When I try to create an array instead, it gives me a generic array creation error. It seems that I would have to make the type less specific like Object[] for this to work.

sgmedev: Oh yeah, Java arrays can get quite funky when it comes to generics. `List.of()` is fine then.

sgmedev (approved):

PR #557, opened 2025-05-10
Crop narration fix

### LINKED ISSUE #546: Growth progress of crops is not narrated

sgmedev:
Closes #546

# Changelog
## Feature Updates
- Crop growth is now narrated as a percentage instead of ripeness


Molitvan:
@TheSuperGamer20578 if you actually looked at #547 before closing it, you'd notice that instead of manually checking for vanilla crops that don't extend CropBlock (like you're doing here), I was checking for other classes/interfaces they maybe extending/implementing (e.g. VegetationBlock, BonemealableBlock) then determining if they are crops with growth stages and getting their growth status using BlockState properties without hardcoding anything instead. This made it so that it'll work with any future crops added to Minecraft as well as even some modded crops.

common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
```diff
@@ -47,19 +49,24 @@
...
+import java.util.*;
 import java.util.stream.Collectors;
 
 /**
  * Translate input objects to narration text.
  */
 @Slf4j
 public class NarrationUtils {
-    public static final Predicate<BlockState> IS_REDSTONE_WIRE = (BlockState state) -> state.getBlock() instanceof RedStoneWireBlock;
+    final static Map<IntegerProperty, Integer> cropAgeProperties = Map.of(
```
BrailleBennett:
```suggestion
    private static final Map<IntegerProperty, Integer> cropAgeProperties = Map.of(
```

BrailleBennett (dismissed):

BrailleBennett (approved):

PR #562, opened 2025-05-15
Refactor key mappings

BrailleBennett:
This PR aims to end our usage of the KeyStroke, KeyInterval, and KeyUtils classes by just doing what literally every other mod does and using the built in logic and detection from the vanilla game.
# Changelog

## Feature Updates
- All bindable key mappings now use Mojang's built in system, removing much unneeded code from the mod

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
@@ -58,6 +58,7 @@ public final class MainClass {
     public static PlayerStatus playerStatus = null;
     public static PlayerWarnings playerWarnings = null;
     public static POIManager poiManager = null;
+    public static PositionNarrator positionNarrator = null;
     public static XPIndicator xpIndicator = null;
 
     private MainClass() {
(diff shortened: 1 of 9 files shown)


BrailleBennett:
Tests don't compile because a bunch of the testing utils are for classes which no longer exist.

common/src/main/java/org/mcaccess/minecraftaccess/features/CameraControls.java
```diff
@@ -1,55 +1,47 @@
...
 import org.mcaccess.minecraftaccess.utils.condition.Interval;
 import org.mcaccess.minecraftaccess.utils.position.Orientation;
 import org.mcaccess.minecraftaccess.utils.position.PlayerPositionUtils;
-import org.mcaccess.minecraftaccess.utils.system.KeyUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * This feature adds the following key binds to control the camera.<br><br>
- * Key binds and combinations:-<br>
```
BrailleBennett: I'll fix this later. I have no idea what happen here.

PR #569, opened 2025-05-23
Fix some locking handler issues

### LINKED ISSUE #566: Add a message when there is no POI selected for locking
When you first join the world, the object tracker has nothing selected. When you currently try to lock onto anything using the Y key while nothing is selected in the object tracker, the Y key appears to do nothing and no action is performed. Instead of this happening, a message should be narrated that says something like "No POI selected in object tracker".

BrailleBennett:
# Changelog

## New Features
- A message will now be spoken if there is no valid object selected to lock onto in the POI tracker
## Bug Fixes
- Aim assist no longer crashes the game
- It is now much more difficult (hopefully impossible) to lock onto air or have an invalid POI object locked


common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
```diff
@@ -126,52 +128,69 @@ private void lookAtLockedTarget() {
...
-                    aimAssistActive = true;
+            if (hostileEntities != null && !hostileEntities.isEmpty()) {
+                Optional<Entity> closestEntity = hostileEntities.stream()
+                        .min(Comparator.comparingDouble(player::distanceTo));
+
+                if (closestEntity.isPresent()) {
+                    Entity entity = closestEntity.get();
+                    if (lockOnEntity(entity)) {
+                        aimAssistActive = true;
+                    }
```
sgmedev: `closestEntity` is always present because there is already a check that `hostileEntities` is not empty.
```suggestion
                Entity closestEntity = hostileEntities.stream()
                        .min(Comparator.comparingDouble(player::distanceTo))
                        .get();
                if (lockOnEntity(entity)) {
                    aimAssistActive = true;
```

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
```diff
@@ -126,52 +128,69 @@ private void lookAtLockedTarget() {
...
+        // Check if player is using a bow
         if (config.aimAssistEnabled && !aimAssistActive && player.isUsingItem() && player.getUseItem().getItem() instanceof BowItem) {
             List<Entity> hostileEntities = BuiltinEntityPOIGroups.HOSTILE.group.getItems();
-            if (!hostileEntities.isEmpty()) {
-                Entity entity = hostileEntities.stream()
-                        .min(Comparator.comparingDouble(e -> WorldUtils.getClientPlayer().distanceTo(e)))
-                        .get();
-                if (lockOnEntity(entity)) {
-                    aimAssistActive = true;
+            if (hostileEntities != null && !hostileEntities.isEmpty()) {
```
sgmedev: `hostileEntities` can never be null because the `getItems()` methods returns the result of a method annotated `@NotNull`.
```suggestion
            if (!hostileEntities.isEmpty()) {
```

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
```diff
@@ -126,52 +128,69 @@ private void lookAtLockedTarget() {
...
-                if (lastAimAssistCue != 1 || bowState != lastBowState) {
-                    PlayerUtils.playSoundOnPlayer(SoundEvents.NOTE_BLOCK_PLING, config.aimAssistAudioCuesVolume, bowState);
-                    lastAimAssistCue = 1;
+            Vec3 eyePosition = player.getEyePosition();
+            Vec3 targetPosition = PlayerUtils.currentEntityLookingAtPosition;
+
+            if (targetPosition != null) {
+                boolean canSeeTarget = PlayerUtils.isPlayerCanSee(eyePosition, targetPosition, lockedOnEntity);
+
+                if (canSeeTarget) {
```
sgmedev: Why are we extracting this into a variable?
```suggestion
                if (PlayerUtils.isPlayerCanSee(eyePosition, targetPosition, lockedOnEntity)) {
```

sgmedev (changes requested):

common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
```diff
@@ -126,52 +128,64 @@ private void lookAtLockedTarget() {
...
         LocalPlayer player = WorldUtils.getClientPlayer();
+        if (player == null) return;
+
+        // Check if player is using a bow
         if (config.aimAssistEnabled && !aimAssistActive && player.isUsingItem() && player.getUseItem().getItem() instanceof BowItem) {
             List<Entity> hostileEntities = BuiltinEntityPOIGroups.HOSTILE.group.getItems();
             if (!hostileEntities.isEmpty()) {
-                Entity entity = hostileEntities.stream()
-                        .min(Comparator.comparingDouble(e -> WorldUtils.getClientPlayer().distanceTo(e)))
+                Entity closestEntity = hostileEntities.stream()
```
sgmedev: Forgot to suggest reverting this variable name previously.
```suggestion
                Entity entity = hostileEntities.stream()
```

BrailleBennett:
Supersedes #521

sgmedev (approved):

PR #579, opened 2025-06-04
Add held item narration key

BrailleBennett:
# Changelog

## New Features
- Added a new keybind (\` by default) that speaks the main hand item the player is holding, when held in conjunction with an alt key, it will speak the current offhand item.


common/src/main/resources/assets/minecraft_access/lang/en_us.json
```diff
@@ -108,6 +108,7 @@
     "minecraft_access.keys.mouse_simulation.right_key_name": "Right Mouse Simulation",
     "minecraft_access.keys.mouse_simulation.scroll_down_key_name": "Mouse Wheel Scroll Down",
     "minecraft_access.keys.mouse_simulation.scroll_up_key_name": "Mouse Wheel Scroll Up",
+    "minecraft_access.keys.other.narrate_held_item_key_name": "Narrate Held Item",
```
sgmedev: Out of order.

docs/content/keybindings.md
```diff
@@ -131,7 +131,8 @@ See also: [Feature Description]({{% relref "/features#position-narrator" %}}), [
 |------------------------------------|-------------------------------------------------------------------------------|
 | `Left Alt` + `Speak Player Status` | Speak only the conditional statuses of the player like air and frost exposure |
 | `Control` + `Speak Player Status`  | Speak currently active effects                                                |
-
+| `Narrate Held Item`                | \`                                                                            | Narrate the item the player currently holds in their main hand |
```
sgmedev: This row has three columns, but the table only has two. This should also probably be in the previous table.

sgmedev (changes requested):
When narrating with an empty hand or offhand it says `air`, it should probably say empty or something.

sgmedev (approved):

PR #587, opened 2025-06-18
Add MC Publish Metadata to show dependencies as embedded

BrailleBennett:
# Changelog

## Bug Fixes
- Mod managers like Modrinth and CurseForge will no longer install embedded dependency mods separately when installing Minecraft Access

### DIFF
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -3,7 +3,7 @@
   "id": "minecraft_access",
   "version": "${version}",
   "name": "Minecraft Access",
-  "description": "A mod that helps visually impaired players to play Minecraft.",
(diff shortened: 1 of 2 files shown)


BrailleBennett:
One issue I just noticed while closely testing is that NeoForge is saying the compatible versions are 1.21.5 and up, while fabric just correctly says 1.21.5. While this issue was not introduced by this PR, it would be worth fixing in it.

BrailleBennett:
> One issue I just noticed while closely testing is that NeoForge is saying the compatible versions are 1.21.5 and up, while fabric just correctly says 1.21.5. While this issue was not introduced by this PR, it would be worth fixing in it.

This is now fixed.

sgmedev (approved):

PR #600, opened 2025-07-09
Add the official test server to the server list in dev environments

BrailleBennett:
If Balm detects that you are running the game in the dev environment, the official testing server will be added to your server list for you.

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
@@ -12,6 +12,8 @@ import net.blay09.mods.balm.client.BalmClientRegistrars;
 import net.blay09.mods.balm.client.platform.event.callback.ClientTickCallback;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.NarratorStatus;
+import net.minecraft.client.multiplayer.ServerData;
+import net.minecraft.client.multiplayer.ServerList;
 import net.minecraft.resources.Identifier;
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
```diff
@@ -194,13 +213,14 @@ private static void initWorldState(Minecraft client) {
...
-        boolean debugMode = Config.getInstance().debugMode || Balm.platform().isDevelopmentEnvironment();
-        if (debugMode) {
+        if (Config.getInstance().debugMode || Balm.platform().isDevelopmentEnvironment()) {
             if (!log.isDebugEnabled()) {
                 Configurator.setLevel("org.mcaccess.minecraftaccess", Level.DEBUG);
             }
-        } else if (log.isDebugEnabled()) {
-            Configurator.setLevel("org.mcaccess.minecraftaccess", Level.INFO);
+        } else {
+            if (log.isDebugEnabled()) {
```
sgmedev: That is an interesting way of doing an else if.

sgmedev (changes requested):

common/src/main/java/org/mcaccess/minecraftaccess/MainClass.java
BrailleBennett: This is exactly why we have code review. I have no idea what I was doing.

sgmedev (approved):

Issue #596, opened 2025-07-04
Mixin refactor


BrailleBennett:
This mods mixins need to be audited, evaluated, and fixed to follow best practices. Below is a snippet of [conversation from the Fabric discord](https://discord.com/channels/507304429255393322/807617700734042122/1390651288924196884) which I have simplified into bullet points for future reference.
## Key Points
- Unique only works on private members, prefixing obviously works on all members, injector handler methods are automatically prefixed
- Your mixin method should always be named to match your intention, not the vanilla method you're injecting/modifying
- all handler methods should be private
- remember not to overuse Inject specifically, it's very rarely the best choice
- All non-static accessor and invoker names should be prefixed and a notation should be added as the target can no longer be inferred
- the fabric wiki is terrible for mixins i would ignore it
- There is really no reason for a mixin ever to be non abstract

## Thoughts on using ``@Inject``
The golden rules are that you should strongly consider switching injectors if you ever need to:
- capture locals
- copy any vanilla code, even 1 call or however small
- replace anything you don't need to
those are all signs that Inject is not the best choice. the general point is Inject doesn't give you any context from or control over the specific thing you're targeting, whereas most other injectors do. eg if you're injecting at a method call, and you want some of the params from that call, you can't get those with Inject without a mixture of local capture and copying them, so you should switch. same goes for the receiver and return value of whatever you're targeting. and if you want to change something you can't do that with Inject without cancelling the method which will lead to you copying code and causing compat issues.
TLDR if you're Injecting and capturing locals or cancelling you should probably consider changing that

BrailleBennett:
Based on confirmation of previously gathered facts that I got today that:
1. All non-accessor classes should be package private and also abstract.
2. All non-accessor or involker methods should be private if at all possible

I proceeded to begin work on making all mixins follow the outlined recommendations via #631.

BrailleBennett:
I was [advised by two members of the Fabric Discord](https://discord.com/channels/507304429255393322/807617700734042122/1430375509862387732) that separation of nested class mixins was the best solution, and that you should only use the nested class name unless the nested class is too generic.

PR #608, opened 2025-07-20
Chat message sounds

Molitvan:

# Changelog

## New Features
- A sound will now play when sending or receiving chat messages


common/src/main/java/org/mcaccess/minecraftaccess/mixin/ChatComponentMixin.java
```diff
@@ -0,0 +1,24 @@
...
+import net.minecraft.client.GuiMessageTag;
+import net.minecraft.client.gui.components.ChatComponent;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.MessageSignature;
+import net.minecraft.sounds.SoundEvents;
+
+@Mixin(ChatComponent.class)
+public class ChatComponentMixin {
+    @Inject(method = "addMessage(Lnet/minecraft/network/chat/Component;Lnet/minecraft/network/chat/MessageSignature;Lnet/minecraft/client/GuiMessageTag;)V", at = @At("HEAD"))
+    public void addMessageMixin(Component chatComponent, MessageSignature headerSignature, GuiMessageTag tag, CallbackInfo ci) {
```
BrailleBennett: You need to describe what the method does in the title, not just say what you're injecting into.

BrailleBennett (changes requested):

BrailleBennett:
I would suggest doing a playNotifySound and tying this volume to the SoundSource.Ui source. This would get rid of the config choice and string for chat sound volume and would be intuitive, as button click sounds are already tied to this slider.

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ChatComponentMixin.java
```diff
@@ -0,0 +1,27 @@
...
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.MessageSignature;
+import net.minecraft.sounds.SoundEvents;
+import net.minecraft.sounds.SoundSource;
+
+@Mixin(ChatComponent.class)
+public class ChatComponentMixin {
+    // Adds a sound whenever a chat message is sent or received if the option is enabled in the config
+    @Inject(method = "addMessage(Lnet/minecraft/network/chat/Component;Lnet/minecraft/network/chat/MessageSignature;Lnet/minecraft/client/GuiMessageTag;)V", at = @At("HEAD"))
+    public void addMessageMixin(Component chatComponent, MessageSignature headerSignature, GuiMessageTag tag, CallbackInfo ci) {
```
BrailleBennett: Injectors should be private as described in #596

BrailleBennett: Parameters, aside from callback info should be excluded, as this injector is not using any parameters in order to function.

BrailleBennett (changes requested):

common/src/main/java/org/mcaccess/minecraftaccess/mixin/ChatComponentMixin.java
```diff
@@ -0,0 +1,27 @@
...
+import net.minecraft.client.GuiMessageTag;
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.gui.components.ChatComponent;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.MessageSignature;
+import net.minecraft.sounds.SoundEvents;
+import net.minecraft.sounds.SoundSource;
+
+@Mixin(ChatComponent.class)
+public class ChatComponentMixin {
```
BrailleBennett: As per #596, all mixins that are not accessors should be package private abstract classes.
```suggestion
abstract class ChatComponentMixin {
```

BrailleBennett (dismissed):

docs/content/sounds/index.md
```diff
@@ -57,6 +57,12 @@ Sounds can be played by pressing the play button.
 |---------------------------------------------------------------------------------------|---------------------------------------------------------------------|
 | [Health], [hunger], [air], or [frost] reached the configured danger warning threshold | {{< wikiSound "Respawn_anchor_deplete1" >}} Shatter and woosh sound |
 
+### Chat
+
+| Description | Sound |
+|-|-|
+| Plays when sending or recieving a chat message | {{< wikiSound "Successful_hit" >}} Ding sound |
```
sgmedev: This table needs formatting.

BrailleBennett:
@molitvan are you still planning on finishing this soon for 1.11.0? If not we can put it in 1.11.1 instead.

sgmedev (approved):

PR #616, opened 2025-08-10
Removed Architectury Loom

### LINKED ISSUE #469: Clean up and audit gradle files
The gradle files in use right now are from 1.17 Minecraft and have been through two mapping migrations. I feel that these files should be audited to make sure there isn't any extra things that aren't needed anymore, and that there isn't coppied logic across the subprojects. We should think about building a new template based on [the Architectury template generator](https://generate.architectury.dev/). However, this generator seems not to be updated often based on refrences to the old shadow plugin, the old restrictive mojmap license, etc. So it shouldn't be taken as the only source for what the base files should look like.

sgmedev:
Replaced Architectury Loom with each platform's own tooling.


BrailleBennett:
After discussion, it sounds like this will be put on hold until deophuscated Minecraft is generally supported by Fabric and NeoForge tooling.

sgmedev:
Superseded by #731

PR #621, opened 2025-08-17
Fixed usage of translated biome name for indicator comparison

BrailleBennett:
# Changelog

## Bug Fixes
- Automatic biome narration will no longer trigger when the game's language is changed
- If a chunk is not fully loaded in, the automatic biome narration will no longer narrate the default biome for the dimention

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
@@ -70,7 +70,6 @@ import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.block.state.properties.BlockStateProperties;
 import net.minecraft.world.level.block.state.properties.ComparatorMode;
 import net.minecraft.world.level.block.state.properties.IntegerProperty;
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/mcaccess/minecraftaccess/utils/NarrationUtils.java
```diff
@@ -637,4 +627,19 @@ public static String narrateEffect(MobEffectInstance effect) {
...
+     * Gets the translated name from registry entry.
+     *
+     * @param holder the holder's registry entry
+     * @param type the type of holder you want the translated name for
+     * @return the holder's human readable name as an Optional
+     */
+    public static Optional<String> getTranslatedName(Holder<?> holder, String type) {
+        Optional<String> translatedName = holder.unwrapKey().map(key -> I18n.get(key.location().toLanguageKey(type)));
+        if (translatedName.isEmpty()) {
+            log.error("Failed to get a valid translation of the %s name".formatted(type));
```
sgmedev:
```suggestion
            log.error("Failed to get a valid translation of the {} name", type);
```

sgmedev (approved):

PR #630, opened 2025-08-29
Hypixel Skyblock Fixes

khanshoaib3:
# Changelog

## Feature Updates

- Added narration exclusions for hypixel skyblock.
    - specifically the health-defence-mana text is suppressed. This text is also cached and spoken when the player stats keybind is used
    - also the health-area-mana text is suppressed and only area is spoken with the "entered" suffix.

## Bug Fixes

- Fixed crash when joining hypixel skyblock.

## Others

- Moved `ScreenReaderInterface.formatNarration` to `NarrationUtils.removeFormattingCodes`.
- Added i18n entry: `minecraft_access.other.hypixel_skyblock_area_entered`


BrailleBennett:
We're currently trying to get rid of Util classes, so I'm not sure how you can justify a new one with one method in it, and I would strongly advise that you move it to a different class, like NarrationUtils.java

PR #631, opened 2025-08-30
Audit utils

BrailleBennett:
This PR audits existing Utility methods and removes simple or redundant ones.
I also began work on refactoring mixin classes as described in #596 by making all injectors private, and all non-accessor classes package private and abstract.


common/src/main/java/org/mcaccess/minecraftaccess/utils/PlayerUtils.java
```diff
@@ -187,104 +140,42 @@ private static BlockHitResult crosshairFluidTarget(double rayCastDistance) {
...
-        if (effects.isEmpty()) {
-            MainClass.narrate(I18n.get("minecraft_access.effect_narration.no_effects"), true);
-            return;
-        }
-        String narration = effects.stream().map(NarrationUtils::narrateEffect)
-                .collect(Collectors.joining(I18n.get("minecraft_access.other.words_connection")));
-        MainClass.narrate(narration, true);
-    }
-
     public static boolean isPlayerTyping() {
```
sgmedev: This only has a single usage and thus should be moved.

common/src/main/java/org/mcaccess/minecraftaccess/utils/PlayerUtils.java
```diff
@@ -187,104 +140,42 @@ private static BlockHitResult crosshairFluidTarget(double rayCastDistance) {
...
-        Screen currentScreen = Minecraft.getInstance().screen;
+        Screen currentScreen = client.screen;
         return currentScreen != null && (currentScreen.getFocused() instanceof EditBox || currentScreen instanceof KeyBindsScreen);
     }
 
     public static boolean isCreative() {
```
sgmedev: This and the methods after are just wrappers around `==` comparisons or simple method calls and most of them only have a couple usages. Do they really have to be here?

common/src/main/java/org/mcaccess/minecraftaccess/mixin/LivingEntityMixin.java
```diff
@@ -14,29 +15,20 @@
...
-        if (Objects.equals(WorldUtils.getClientPlayer(), this)) {
-            EffectNarrator.narrateGained(effectInstance);
-        }
-    }
-
-    @Inject(method = "onEffectAdded", at = @At("TAIL"))
-    public void narrateEffectApplication2(MobEffectInstance effectInstance, Entity entity, CallbackInfo ci) {
-        if (Objects.equals(WorldUtils.getClientPlayer(), this)) {
+    @Inject(method = {"onEffectUpdated", "onEffectAdded"}, at = @At("TAIL"))
+    private void narrateEffectApplication(MobEffectInstance effectInstance, boolean forced, Entity entity, CallbackInfo ci) {
```
sgmedev, line 23: `onEffectAdded` doesn't have a `forced` parameter and from a quick two minutes of experimentation with the export thingo just now it looks like if you have parameters and they don't perfectly match Mixin will *silently* not inject (amazing) even if your parameters are a subset of the actual method, but not having any parameters other than `CallbackInfo` is perfectly valid and will inject according to the JavaDoc of `@Inject`. I wonder if Mixin Extras has anything more powerful?

sgmedev (changes requested):
This introduces several NullPointerException warnings in IntelliJ. Please check weather any of these are of concern and add assertions or otherwise as appropriate.

In `PlayerPositionUtils` the following methods should be reconsidered:
- `getX()`, `getY()`, and `getZ()` — Single usage, simple field access
- `getPlayerBlockPosition()` — Unused
- `getPlayerPosition()` — Looks cool until you consider the only time the optional is actually used is in the unused `getPlayerBlockPosition()` and everywhere else `.orElseThrow()` is called making it a simple field access with extra steps, it also only has two usages outside of the methods above

`WorldUtils` only has a singular six line method which only has a single usage. This should become a private method in the class using it. It also has a few unused imports which reviewdog filtered out because they aren't technically on your diff because they were there before.

`MouseUtils.move()` and `MouseUtils.performAt()` are also unused.

BrailleBennett:
For game mode stuff, I feel like it's worth having them for simplicity, as those are very universal things we call across the code base
It's nice to be able to reach for a specific game mode check without having to realy dig into how to make it work
isTyping I made a method just because I figured there were future applications that may exist, but I could probably see how that one could be changed to just be inline.
I didn't realize WorldUtils only had one method left though, I'll definitely move and delete that class.

sgmedev:
> isTyping I made a method just because I figured there were future applications that may exist

This is the thinking that led to utils getting out of hand in the first place.

> For game mode stuff, I feel like it's worth having them for simplicity

They are literally an `==` comparison, and in most places we already have the client object.

> as those are very universal things we call across the code base

It's not actually, creative, spectator, and hardcore have an entire two usages each and `isPlayerSpectating()` has three usages. Adventure and survival have four and five usages each, respectively, but this is still far from universal, especially when you consider that these usages are localised to just two classes.

sgmedev (approved):

PR #634, opened 2025-08-31
Feature/narrate bossbars

BrailleBennett:
# Changelog

## New Features
- The U key now narrates bossbars in a cycling list, including those added via a server or adventure map
## Feature Updates
- The advancements key is now bound to O by default, instead of being unbound


BrailleBennett:
I chose to bind the advancements key by default because there are some low vision players to which this key is still useful, additionally, I thought it would make sense to be bound, as I almost accidentally put the bossbar narration key on it, even though we previously decided that we'd be binding the advancements key there.
Additionally, I didn't add an automatic bossbar narration mechanism, but I'm not sure if we need one. Let me know your thoughts on that.

common/src/main/java/org/mcaccess/minecraftaccess/features/HUDStatus.java
```diff
@@ -1,19 +1,54 @@
...
         }
+
+        if (KeyBindingsHandler.NARRATE_BOSSBARS_KEY.mapping.consumeClick()) {
+            if (!bossbarKeyIsDown) {
+                bossbarKeyIsDown = true;
+                narrateBossBars();
+            }
+        } else if (!KeyBindingsHandler.NARRATE_BOSSBARS_KEY.mapping.isDown()) {
+            bossbarKeyIsDown = false;
+        }
```
sgmedev, line 35: This will need to be revisited in #562.

sgmedev (approved):

PR #651, opened 2025-10-19
Feature/durability warnings

### LINKED ISSUE #589: Add low durability warning for held items
From [Logic](https://youtube.com/@logicproxgaming),
Add a configurable warning, similar to health, hunger, etc., that triggers when your tools durability in the main hand or offhand reaches a configured percentage value. Additionally, have an optional feature that also notifies you when your worn armor is low on durability as well.

BrailleBennett:
This PR refactors the main tick method of the mod, take great care when reviewing this part of the PR. I'm confident I did it right, and I tested, but we can't be too careful.

# Changelog

## New Features
- Added optional warnings for armor and tools when the durability falls below the two configured percentages

## Feature Updates
- If a warning has two thresholds, as is the case for durability and health, the more severe threshold will play a metal clang sound instead of the normal shatter whoosh sound.


BrailleBennett:
I need to do documentation

sgmedev (approved):

PR #653, opened 2025-10-20
Prevent air and dying entities in the object tracker at all costs

### LINKED ISSUE #583: Expand the check for if the locked on entity/block is valid to everywhere in the object tracker
Currently it only covers locking, but it should also cover narrating the current object in object tracker and targeting the nearest object.

BrailleBennett:
# Changelog

## Bug Fixes
- Air and dying entities will now no longer appear in the Object Tracker.


BrailleBennett:
Adding @Molitvan because I'd like their thoughts on what I did to the Object Tracker and if my changes make sense to them even.

sgmedev (approved):

BrailleBennett:
I'll give @Molitvan until tomorrow afternoon CT to review, I'd still like to hear from them and what they think of the changes, but I do understand they might be busy, so I don't want to hold up the release. We can always do a revision PR later if nothing else.

PR #665, opened 2025-11-06
Switch Architectury to Balm

BrailleBennett:
# Changelog

## Others
- Minecraft Access now uses Balm instead of Architectury API, this should hopefully lead to increase stability and faster version updates, though cloth Config is not controlled by the Balm devs, and is another dependency

### DIFF
--- a/fabric/src/main/java/org/mcaccess/minecraftaccess/fabric/MainClassFabric.java
+++ b/fabric/src/main/java/org/mcaccess/minecraftaccess/fabric/MainClassFabric.java
@@ -1,6 +1,5 @@
 package org.mcaccess.minecraftaccess.fabric;
 
-import net.blay09.mods.balm.api.Balm;
 import net.blay09.mods.balm.api.EmptyLoadContext;
 import net.blay09.mods.balm.api.client.BalmClient;
 import net.fabricmc.api.ModInitializer;
@@ -10,7 +9,6 @@ import org.mcaccess.minecraftaccess.MainClass;
 public class MainClassFabric implements ModInitializer {
     @Override
     public void onInitialize() {
-        Balm.initializeMod(MainClass.MOD_ID, EmptyLoadContext.INSTANCE, () -> {});
         BalmClient.initializeMod(MainClass.MOD_ID, EmptyLoadContext.INSTANCE, MainClass::init);
     }
 }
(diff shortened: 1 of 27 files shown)


fabric/src/main/java/org/mcaccess/minecraftaccess/fabric/MainClassFabric.java
```diff
@@ -1,12 +1,16 @@
...
+import net.blay09.mods.balm.api.client.BalmClient;
 import net.fabricmc.api.ModInitializer;
 
 import org.mcaccess.minecraftaccess.MainClass;
 
 public class MainClassFabric implements ModInitializer {
     @Override
     public void onInitialize() {
-        MainClass.init();
+        Balm.initializeMod(MainClass.MOD_ID, EmptyLoadContext.INSTANCE, () -> {});
```
sgmedev: The documentation only says you need the client initialiser in the client entrypoint, this common initialiser appears to do absolutely nothing. Same with NeoForge.

https://balm.twelveiterations.com/docs/concepts/entrypoints#client-entrypoint-1

neoforge/src/main/java/org/mcaccess/minecraftaccess/neoforge/MainClassNeoForge.java
```diff
@@ -17,10 +18,8 @@ public class MainClassNeoForge {
     public MainClassNeoForge(IEventBus modEventBus, ModContainer container) {
         modEventBus.register(this);
         container.registerExtensionPoint(IConfigScreenFactory.class, (modContainer, parent) -> AutoConfig.getConfigScreen(Config.class, parent).get());
-    }
-
-    @SubscribeEvent
-    private void onClientSetup(final FMLClientSetupEvent event) {
-        MainClass.init();
+        final NeoForgeLoadContext context = new NeoForgeLoadContext(modEventBus);
```
sgmedev: We do not currently use immutable local variables anywhere and this variable is not special. If you want to adopt that as a standard it should be subject to proper discussion and a refactor to apply it everywhere instead of just one random variable somewhere.

sgmedev (changes requested):

sgmedev (approved):

PR #708, opened 2026-01-26
Change locking sound config

### LINKED ISSUE #582: Play Sound option in Locking settings is mislabeled
It's labeled as just p"Play sound" and doesn't mention that you can make it speak instead.

A better label would be something like "Unlock notification" with the modes being "Sound" and "Speech" instead of yes and no.

BrailleBennett:
# Changelog

## Feature Updates
- Removed unlocking sound config
- Made picked up item narration one option instead of having fishing harvest be a separate option


common/src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/LockingHandler.java
```diff
@@ -225,12 +225,7 @@ private void unlock(boolean narrate, boolean isStillValid) {
         if (!isStillValid) MainClass.poiManager.objectTracker.clearCurrentObject();
 
         if (narrate) {
-            if (config.unlockingSound) {
-                assert Minecraft.getInstance().player != null;
-                Minecraft.getInstance().player.playSound(SoundEvents.NOTE_BLOCK_BASEDRUM.value(), 0.4f, 2.0f);
-            } else {
                 MainClass.narrate(I18n.get("narrator.button.difficulty_lock.unlocked"), true);
```
sgmedev: Double indentation.

sgmedev (changes requested):

sgmedev (approved):

PR #601, opened 2025-07-10
Refactor code wording for consistency

BrailleBennett:
this PR is entirely code facing (besides the docs grammarly fix up). It aims to unify previously fragmented terminology to follow the phrasing used by the official mappings if available. For example all "speak", "read", "notify", etc. words have been consolidated to use the word narrate. All methods that execute each tick are renamed to tick, and some other wording was polished to provide code clarity and a consistent set of vocabulary project-wide.

PR #724, opened 2026-03-11
Make narration hints and crouch&sprint sounds configurable

### LINKED ISSUE #717: Add config for crouch and sprint cues
Currently, the crouching and sprinting hints that play when the player does those actions are not configurable, as some players might not want or need this information, this should definitely be configurable and should be trivial to add.

BrailleBennett:
# Changelog

## New Features
- Added config for narrating the built in Minecraft narration hints involving interaction and positional hints (defaulted to enabled)

## Feature Updates
- Added config for the playing of the crouch and sprint sounds previously added (defaulted to enabled)

### DIFF
--- a/common/src/main/java/org/mcaccess/minecraftaccess/Config.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/Config.java
@@ -96,7 +96,7 @@ public final class Config implements ConfigData {
 
     public static final class SpeechSettings {
         public float speechRate = 50;
-        public boolean speakHints = true;
+        public boolean narrateHints = true;
 
         private SpeechSettings() {
         }
--- a/common/src/main/java/org/mcaccess/minecraftaccess/Config.java
+++ b/common/src/main/java/org/mcaccess/minecraftaccess/Config.java
@@ -78,6 +78,7 @@ public final class Config implements ConfigData {
         public boolean timeIndicatorEnabled = true;
         public boolean xpIndicatorEnabled = true;
         public boolean facingDirectionEnabled = true;
+        public boolean crouchAndSprintCues = true;
         @ConfigEntry.Gui.EnumHandler(option = ConfigEntry.Gui.EnumHandler.EnumDisplayOption.BUTTON)
         public PickedUpItemNarration pickedUpItemNarration = PickedUpItemNarration.WHEN_FISHING;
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/mcaccess/minecraftaccess/features/PlayerStatus.java
```diff
@@ -78,15 +78,15 @@ public void initialize() {
...
                         return;
                     }
                     level.playPlayerSound(SoundEvents.SHOVEL_FLATTEN, SoundSource.PLAYERS, 1.0f, value ? 0.5f : 0.9f);
                 }
         );
 
         new ServerChangeDetector<Boolean>().levelEvent(
-                (client, player, level) -> player.isSprinting() && !player.isCrouching(),
+                (client, player, level) -> Config.getInstance().features.crouchAndSprintCues && player.isSprinting() && !player.isCrouching(),
                 (client, player, level, previous, value) -> level.playPlayerSound(SoundEvents.SHOVEL_FLATTEN, SoundSource.PLAYERS, 1.0f, value ? 2.5f : 0.9f)
```
sgmedev: The config should be checked in the event handler as it is correctly done above. Checking it in the first parameter means that it's assuming that the player is always not sprinting when the config is disabled which theoretically means that the event will fire whenever the config is changed if the player is sprinting.
```suggestion
                (client, player, level) -> player.isSprinting() && !player.isCrouching(),
                (client, player, level, previous, value) -> {
                    if (Config.getInstance().features.crouchAndSprintCues) {
                        level.playPlayerSound(SoundEvents.SHOVEL_FLATTEN, SoundSource.PLAYERS, 1.0f, value ? 2.5f : 0.9f)
                    }
                }
```

common/src/main/java/org/mcaccess/minecraftaccess/Config.java
```diff
@@ -95,6 +96,7 @@ public enum PickedUpItemNarration {
 
     public static final class SpeechSettings {
         public float speechRate = 50;
+        public boolean speakHints = true;
```
sgmedev: This should be `narrateHints` as per #601.
```suggestion
        public boolean narrateHints = true;
```

sgmedev (changes requested):

sgmedev (approved):

PR #727, opened 2026-03-17
Transition to built in Balm SessionLocals

BrailleBennett:
This PR changes our home made ServerLocal implementation to use BALM's new SessionLocal one. There should be 0 player facing change. Additionally, it adds more values under this SessionLocal type instead of having event hooks for on world join.


BrailleBennett:
I believe I got them all, you may want to do a search for the join hook registration to make sure though.

BrailleBennett:
Since we're most likely going to release this build as the final 1.21.11 snapshot, I felt it was appropriate to bump BALM due to the bugs discovered in the 21.11.7 build.

sgmedev (approved):

PR #731, opened 2026-04-25
Minecraft 26.1

### LINKED ISSUE #469: Clean up and audit gradle files

sgmedev:
Upgraded to Minecraft 26.1.2
# Changelog

## New Features
- Both Fabric and NeoForge now share a single mod jar file, reducing complexity


BrailleBennett:
Remind me to follow up on everyone's maven preferences before finalizing this. I think I have it correct, but I will triple check. @sgmedev

settings.gradle
```diff
@@ -49,8 +32,9 @@ pluginManagement {
 }
 
 plugins {
-    id 'com.gradle.develocity' version '4.3.2'
-    id 'dev.architectury.loom' version '1.13.467'
+    id 'com.gradle.develocity' version '4.4.0'
+    id 'net.fabricmc.fabric-loom-repositories' version '1.16-SNAPSHOT'
```
sgmedev: Remember to pin before merging

BrailleBennett:
Final remaining issues:
- Tolk bindings cause NeoForge not to run in the dev environment due to out of transformation area mixin attempting to be applied (will be fixed by either applying an artifact transformation to Tolk to make it a mod or by using prism for TTS in the future)
- Jar in Jarring appears to have issues and doesn't appear to be working for either Fabric or NeoForge right now

BrailleBennett (dismissed):

BrailleBennett (approved):

PR #747, opened 2026-06-20
Update mod to 26.2

BrailleBennett:
# Changelog

## New Features
- The mod now runs on game version 26.2

## Feature Updates
- The advancements key is now bound to X by default so as not to conflict with the friends list key (O by default)

## Bug Fixes
- Slimes, Magma Cubes, Shulkers, and Ghasts are all now properly recognized as hostile enemies in the Object Tracker


src/main/java/org/mcaccess/minecraftaccess/addon/accessmenu/Weather.java
```diff
@@ -23,29 +23,26 @@ public void execute() {
         if (Minecraft.getInstance().player == null) return;
         if (level == null) return;
         if (currentBiome == null) return;
+        Minecraft.getInstance().player.clientSideCloseContainer();
```
sgmedev: Is this actually needed at all? When refactoring the access menu I made the GUI close itself when calling a function instead of the functions each being responsible for closing the current (presumable access menu) screen. It is possible that this is just one that I missed.

src/main/java/org/mcaccess/minecraftaccess/compat/mixin/clothconfig/ScreenNarrationCollectorOutputMixin.java
```diff
@@ -23,12 +23,12 @@ abstract class ScreenNarrationCollectorOutputMixin {
...
 
-    @ModifyVariable(method = "add", at = @At("HEAD"), argsOnly = true)
+    @ModifyVariable(method = "add", at = @At("HEAD"), argsOnly = true, name = "contents")
     private NarrationThunk<?> makeSameOptionValuesBeingNarrated(NarrationThunk<?> contents) {
-        boolean clothScreenOpening = Minecraft.getInstance().screen instanceof ClothConfigScreen;
+        boolean clothScreenOpening = Minecraft.getInstance().gui.screen() instanceof ClothConfigScreen;
         boolean mightBeValueDepth = depth > 1;
         if (clothScreenOpening && mightBeValueDepth && ((NarrationThunkAccessor) contents).getContents() instanceof String content) {
-            String prefix = IntStream.range(0, whiteSpaceCount).mapToObj(i -> " ").reduce("", String::concat);
+            String prefix = IntStream.range(0, whiteSpaceCount).mapToObj(_ -> " ").reduce("", String::concat);
```
sgmedev:
```java
" ".repeat(whiteSpaceCount)
```

src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/ObjectTracker.java
```diff
@@ -266,27 +267,24 @@ private void moveGroup(int step) {
             return;
         }
 
-        currentObject = validObjects.getFirst();
```
sgmedev, line 269: Why is this no longer needed?

src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/ObjectTracker.java
```diff
@@ -266,27 +267,24 @@ private void moveGroup(int step) {
...
-        if (object == null) return false;
-
-        if (object instanceof Entity entity) {
-            return entity.isAlive();
-        }
-
-        if (object instanceof BlockPos pos) {
-            if (Minecraft.getInstance().level == null) return false;
-            return !(Minecraft.getInstance().level.getBlockState(pos).getBlock() instanceof AirBlock);
+        switch (object) {
```
sgmedev: It might be better to use a switch expression and return it so that each branch is statically required to yield something instead of having default fall through to the outer return statement.

src/main/java/org/mcaccess/minecraftaccess/utils/PlayerUtils.java
```diff
@@ -84,11 +88,18 @@ public static void lookAt(Entity entity) {
...
-        return hitResult.getType() == HitResult.Type.MISS;
+        try (Level level = entity.level()) {
+            BlockHitResult hitResult = level.clip(
+                    new ClipContext(somewhereOnEntity, playerEyePos,
+                            ClipContext.Block.COLLIDER,
+                            ClipContext.Fluid.NONE, entity));
+            return hitResult.getType() == HitResult.Type.MISS;
+        } catch (IOException e) {
+            log.error("IoException in PlayerUtils from isVissibleToPlayer");
+            e.printStackTrace();
```
sgmedev: The exception can be passed to `log.error` directly which will properly log it with the error as opposed to directly outputting it to stderr bypassing the logger.

src/main/java/org/mcaccess/minecraftaccess/utils/PlayerUtils.java
```diff
@@ -84,11 +88,18 @@ public static void lookAt(Entity entity) {
     }
 
     public static boolean isVisibleToPlayer(Vec3 playerEyePos, Vec3 somewhereOnEntity, Entity entity) {
-        BlockHitResult hitResult = entity.level().clip(
-                new ClipContext(somewhereOnEntity, playerEyePos,
-                        ClipContext.Block.COLLIDER,
-                        ClipContext.Fluid.NONE, entity));
-        return hitResult.getType() == HitResult.Type.MISS;
+        try (Level level = entity.level()) {
```
sgmedev: This means that `level.close()` will be called when exiting this block which I don't think is what we want. `entity.level()` is a simple getter which returns the value of the entity's `level` field, nothing is "opened" which we would be responsible for closing.

src/main/java/org/mcaccess/minecraftaccess/MainClass.java
```diff
@@ -84,13 +84,11 @@ public static void init(BalmClientRegistrars registrars, List<Addon> addons) {
...
         new AutoLibrarySetup().initialize();
 
-        ClientPlayingTick.AFTER.configureMapping((priority, callback) -> {
-            ClientTickCallback.AFTER.register(priority, client -> {
-                if (client.player != null && client.level != null) {
-                    callback.handle(client, client.player, client.level);
-                }
-            });
-        });
+        ClientPlayingTick.AFTER.configureMapping((priority, callback) -> ClientTickCallback.AFTER.register(priority, client -> {
```
sgmedev: This was more readable before, it might be worth switching this inspection to a suggestion only because there are lots of cases where it doesn't make much sense. Alternatively, we could still wrap it to a new line and forgo braces, but I'm not entirely sure which is better. I think I also saw some other instances of this, though they were less bad.

src/main/java/org/mcaccess/minecraftaccess/mixin/GameModeSwitcherScreenMixin.java
```diff
@@ -19,7 +19,7 @@ abstract class GameModeSwitcherScreenMixin {
     private GameModeIconAccessor previous;
 
     @WrapOperation(
-            method = {"init", "keyPressed", "render"},
+            method = {"init", "keyPressed"},
```
sgmedev: This breaks narration when hovering with the mouse. I think they renamed rendering to extraction for some reason.

sgmedev (changes requested):

src/main/java/org/mcaccess/minecraftaccess/compat/mixin/clothconfig/ScreenNarrationCollectorOutputMixin.java
BrailleBennett: What do you mean by this?

src/main/java/org/mcaccess/minecraftaccess/features/point_of_interest/ObjectTracker.java
BrailleBennett: Great call.

src/main/java/org/mcaccess/minecraftaccess/utils/PlayerUtils.java
BrailleBennett: I was confused because it implements AutoClosable, but I can just ignore it.

src/main/java/org/mcaccess/minecraftaccess/mixin/GameModeSwitcherScreenMixin.java
BrailleBennett: Yeah, I definitely should have done more research before just removing it with no replacement. Thanks for catching it. The closest thing I found was extractRenderState, but that seems to do with mouse handling, so I think it's the right method?

src/main/java/org/mcaccess/minecraftaccess/utils/PlayerUtils.java
BrailleBennett: TIL! This will be good to know in future.

sgmedev (dismissed):
Just need to fix the Checkstyle errors and test failure.

sgmedev (approved):

PR #762, opened 2026-07-24
Balm config

### LINKED ISSUE #510: Array config options are inaccessible
Array config options are not tab selectable at all and probably require further work to make them usable.

Currently the only array config option is `readCrosshair.partialSpeaking.targets` and this wasn't in the config menu at all before #347, therefore this is low priority and not a regression in #347.

### LINKED ISSUE #709: Cloth Config I18n Keys have a - instead of _
The text i18n keys that make up the cloth config screen E.G. text.minecraft-access.features.pickedUpItemNarration use a hyphen in minecraft-access instead of an _ like is proper when spacing keys by our mod ID, as the ID itself is minecraft_access.

BrailleBennett:
# Changelog
## New Features
- The mod now utilizes the built in BALM config Screen
- No longer bundle Cloth Config in the mod jar


BrailleBennett:
# TODO
- Have @sgmedev port the custom cloth screen to BALM's system
- Change I18n keys to match new format of Balm Config Screen
- Double check that Modmenu is runtime only
- Double check that Cloth Config common is only in the dev environment but not at runtime or bundled in the jar
- Figure out if Cloth should be clothConfig or clothConfigCommon in the versions catalog

This branch will currently only run if you supply the cloth config jar in the mods folder yourself before running the client.

PR #764, opened 2026-07-27
Implement narrator fallback on unsupported platforms

BrailleBennett:
# Changelog

## New Features
- If your platform isn't supported by our speech solutions, the mod will attempt to let vanilla handle narration instead

### DIFF
--- a/src/main/java/org/mcaccess/minecraftaccess/features/AutoLibrarySetup.java
+++ b/src/main/java/org/mcaccess/minecraftaccess/features/AutoLibrarySetup.java
@@ -34,8 +34,8 @@ public class AutoLibrarySetup {
      * Installs <a href="https://github.com/ndarilek/tolk">tolk</a> for windows and <a href="https://github.com/khanshoaib3/libspeechdwrapper">libspeechdwrapper</a> for linux.
      */
     private void downloadAndInstall() throws IOException {
-        switch (Util.getPlatform()) {
-            case Util.OS os when os == Util.OS.WINDOWS && isAmd64() -> {
+        switch (getSupportedLibArchitecture()) {
+            case WINDOWS -> {
                 log.info("Downloading latest tolk build...");
                 File tolkLatestBuildZip = new File(Paths.get("tolk-latest-build.zip").toAbsolutePath().toString());
                 FileUtils.copyURLToFile(URI.create("https://github.com/ndarilek/tolk/releases/download/refs%2Fheads%2Fmaster/tolk.zip").toURL(), tolkLatestBuildZip);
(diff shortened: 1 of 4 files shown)


src/main/java/org/mcaccess/minecraftaccess/features/AutoLibrarySetup.java
```diff
@@ -37,7 +35,7 @@ public void initialize() {
      */
     private void downloadAndInstall() throws IOException {
         switch (Util.getPlatform()) {
```
sgmedev: These switch statements are turning into a weird else-if chain. If implementing it this way it might be better to just use an if-else, however it also looks like the `isAmd64()` check could just be extracted out.

src/main/java/org/mcaccess/minecraftaccess/features/AutoLibrarySetup.java
```diff
@@ -98,7 +96,7 @@ private boolean checkInstalled() {
     private List<String> getRequiredLibraryNames() {
         List<String> requiredFiles = new ArrayList<>();
         switch (Util.getPlatform()) {
-            case WINDOWS -> {
+            case Util.OS os when os == Util.OS.WINDOWS && isAmd64() -> {
                 requiredFiles.add("Tolk.dll");
                 if (System.getProperty("os.arch").equalsIgnoreCase("X86")) {
```
sgmedev, line 101: If 64-bit && 32-bit. Should probably include `x86` in the list of supported architectures.

sgmedev (changes requested):

src/main/java/org/mcaccess/minecraftaccess/mixin/GameNarratorMixin.java
```diff
@@ -11,19 +13,28 @@
...
 
     @Inject(method = "narrateMessage", at = @At("HEAD"), cancellable = true)
     private void narrateMessage(String message, boolean interrupt, CallbackInfo ci) {
-        if (MainClass.getScreenReader() == null || !MainClass.getScreenReader().isInitialized()) {
-            return;
+        if (MainClass.getScreenReader() != null && MainClass.getScreenReader().isInitialized()) {
+            MainClass.getScreenReader().narrate(message, interrupt);
+            ci.cancel();
+        } else if (!hasWarned) {
+            log.error("Failed to narrate using the Minecraft Access logic. Will fallback to vanilla handling.");
```
sgmedev: I wonder if this could be moved to where it is initialised to save the whole `hasWarned` thing.

BrailleBennett:
This solution feels differently cursed to me. It uses an Enum that Mojang made instead of our own to save on code duplication, but that also might make it more confusing to read.jj

BrailleBennett:
This will be resolved by the implementation of prism

PR #771, opened 2026-08-16
Narrate carried items in inventory screens

### LINKED ISSUE #388: Narrate Held Item key should read held item in inventory screen
- [ ] Make the player status key (R by default) read the item that's currently held by the cursor when an inventory screen is open
- [ ] #389

### LINKED ISSUE #395: Light blocks in creative inventory and hotbar do not narrate light level
In the OP items tab in the creative inventory (togglable from the controls menu in creative) the light blocks do not say what light level they are. This issue seems similar to the music disc issue previously fixed where all music discs didn't report their name.

BrailleBennett:
# Changelog

## New Features
- Using the narrate held item key in your inventory now speaks items that are carried by your cursor
- Light block light level is now narrated in the inventory.
- Test block mode is now narrated in the inventory.
- Bundle fullness progress is now narrated in the inventory.
- item narration is now shared across the entire mod for consistency

### DIFF
--- a/src/main/java/org/mcaccess/minecraftaccess/features/NarrateHeldItem.java
+++ b/src/main/java/org/mcaccess/minecraftaccess/features/NarrateHeldItem.java
@@ -89,7 +89,7 @@ public class NarrateHeldItem implements BalmClientModule {
         previousSelectedSlot.value = selectedSlot;
     }
 
-    private String getItemName(ItemStack itemStack, boolean addCount) {
+    public static String getItemName(ItemStack itemStack, boolean addCount) {
         if (itemStack.isEmpty()) {
             return I18n.get("minecraft_access.inventory_controls.empty_slot", "");
         }
--- a/src/main/java/org/mcaccess/minecraftaccess/features/NarrateHeldItem.java
+++ b/src/main/java/org/mcaccess/minecraftaccess/features/NarrateHeldItem.java
@@ -44,16 +44,9 @@ public class NarrateHeldItem implements BalmClientModule {
                 .withDefault(InputBinding.key(InputConstants.KEY_GRAVE))
                 .overrideCategory(KeyMappingCategories.OTHER)
                 .handleScreenInput(event -> {
-                    if (event.screen() instanceof AbstractContainerScreen) {
-                        ItemStack heldItem = Minecraft.getInstance().player.containerMenu.getCarried();
-                        String heldItemName = getItemName(heldItem);
-                        int heldItemCount = heldItem.getCount();
-                        heldItemName = (heldItemCount != 1 && !heldItem.isEmpty()) ? heldItemCount + " " + heldItemName : heldItemName;
-
-                        MainClass.narrate(heldItemName, false);
-                        return true;
-                    }
-                    return false;
+                    if (!(event.screen() instanceof AbstractContainerScreen)) return false;
+                    MainClass.narrate(getItemName(Minecraft.getInstance().player.containerMenu.getCarried(), true), false);
+                    return true;
                 })
                 .handleWorldInput(_ -> {
                     narrateHand(false);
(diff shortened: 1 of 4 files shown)


BrailleBennett:
Two concerns
- Should there be a prefix like "holding [item]" or "carrying [item",
- should the item narration be abstracted into a helper function as it's basically copied straight from the narrate hand portion of the class?

sgmedev:
> * should the item narration be abstracted into a helper function as it's basically copied straight from the narrate hand portion of the class?

That's probably a good idea, it does feel like just enough logic for someone to forget about it and cause divergence.

src/main/java/org/mcaccess/minecraftaccess/features/NarrateHeldItem.java
```diff
@@ -42,6 +43,18 @@ public void initialize() {
         Kuma.createKeyMapping(Identifier.fromNamespaceAndPath(MainClass.MOD_ID, "other.narrate_held_item/mainhand"))
                 .withDefault(InputBinding.key(InputConstants.KEY_GRAVE))
                 .overrideCategory(KeyMappingCategories.OTHER)
+                .handleScreenInput(event -> {
+                    if (event.screen() instanceof AbstractContainerScreen) {
```
sgmedev: Nit: It might be a good idea to invert this if statement to reduce nesting.

src/main/java/org/mcaccess/minecraftaccess/features/NarrateHeldItem.java
```diff
@@ -63,27 +75,27 @@ private void tick(Minecraft client, Player player, Level level) {
...
         boolean slotChanged = !Objects.equals(selectedSlot, previousSelectedSlot.value);
 
         if (nameChanged || slotChanged) {
-            MainClass.narrate(I18n.get("minecraft_access.other.selected", itemNameWithCount), true);
-        } else if (countChanged && Config.getInstance().features.narrateHeldItemsCountWhenChanged) {
+            MainClass.narrate(I18n.get("minecraft_access.other.selected", getItemName(currentItemStack, true)), true);
+        } else if (countChanged) {
             MainClass.narrate(String.valueOf(itemCount), true);
+        } else {
+            return;
```
sgmedev: This probably shouldn't return early out of principal, if for some reason this if statement becomes bugged then the previous fields won't be updated. I also question why this isn't using the `ChangeDetector` class which is designed exactly for cases like this though that's probably a bit out of scope and there might be a reason unless I just missed this when I did that refactor.

src/main/java/org/mcaccess/minecraftaccess/features/NarrateHeldItem.java
```diff
@@ -63,27 +75,27 @@ private void tick(Minecraft client, Player player, Level level) {
...
         int selectedSlot = player.getInventory().getSelectedSlot();
-        String baseItemName = getItemName(currentItemStack);
+        String baseItemName = getItemName(currentItemStack, false);
         int itemCount = currentItemStack.getCount();
 
-        String itemNameWithCount = (currentItemStack.getCount() != 1 && !currentItemStack.isEmpty()) ? itemCount + " " + baseItemName : baseItemName;
-
         boolean nameChanged = !Objects.equals(baseItemName, previousItemName.value);
-        boolean countChanged = !Objects.equals(itemCount, previousItemCount.value);
+        boolean countChanged = Config.getInstance().features.narrateHeldItemsCountWhenChanged && !Objects.equals(itemCount, previousItemCount.value);
```
sgmedev: This is conflating the change with whether it should act on the change. The if statement below now reads as "narrate if only the count changed" as opposed to "narrate if only the count changed and the user has that setting enabled". You could rename the variable to work around this but I feel like having the change detection have its own statement is probably a bit more readable and there isn't anything wrong with using a config option in an if statement.

src/main/java/org/mcaccess/minecraftaccess/features/NarrateHeldItem.java
```diff
@@ -95,30 +107,43 @@ private String getItemName(ItemStack itemStack) {
...
+        if (itemStack.is(Items.TEST_BLOCK)) {
+            TestBlockMode mode = Optional.ofNullable(itemStack.get(DataComponents.BLOCK_STATE))
+                    .map(blockState -> blockState.get(TestBlock.MODE))
+                    .orElse(TestBlockMode.START);
+            itemName.append(' ').append(mode.getDisplayName().getString());
+        }
+
+        if (itemStack.has(DataComponents.BUNDLE_CONTENTS)) {
+            int weight = Math.round(BundleItem.getFullnessDisplay(itemStack) * 64);
+            itemName.append(' ').append(weight).append("/64");
```
sgmedev: It might be better to use `BundleContents` directly. It looks like it exposes the current weight as an int-based `Fraction` meaning floats wouldn't be involved at all. Alternatively, I wonder if using it as a percentage may be a good idea?

sgmedev (changes requested):

PR #772, opened 2026-08-16
Wolf collars

### LINKED ISSUE #750: Tamed Wolf Collar Color Not Reported
When looking at a tamed wolf, it has a colorable collar that the player can change by interacting on the wolf with a dye item. The color and change of this collar is not reported in crosshair narration or the object tracker list.

BrailleBennett:
# Changelog

## New Features
- The collars which tamed cats and wolfs wear are now narrated as equipment instead of being ignored
## Feature Updates
- The mod now uses translations for custom entity poses added by other mods if defined properly when narrating

### DIFF
--- a/src/main/java/org/mcaccess/minecraftaccess/addon/worldnarrators/MinecraftAccess.java
+++ b/src/main/java/org/mcaccess/minecraftaccess/addon/worldnarrators/MinecraftAccess.java
@@ -2,6 +2,7 @@ package org.mcaccess.minecraftaccess.addon.worldnarrators;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


BrailleBennett:
I'm not sure if the entity pose switch should be converted to a similar structure to what we do for moon phases, where the key is just retrieved without separate branch logic for each enum value.

sgmedev:
> I'm not sure if the entity pose switch should be converted to a similar structure to what we do for moon phases, where the key is just retrieved without separate branch logic for each enum value.

It looks like that would be a good idea, it does feel a bit weird how you've now got a bunch of cases appending the pose name on to the base key and then a generic default which does the same thing but more generic. Is there any discrepancy between the key that your new default branch would use and the specific branches or could you cleanly delete those lines and have the same behaviour?

src/main/java/org/mcaccess/minecraftaccess/addon/worldnarrators/MinecraftAccess.java
```diff
@@ -280,8 +297,8 @@ private static String getSheepInfo(Sheep sheep, String currentQuery) {
      * @param blockPos block position (in the client world)
      * @param side     if side is provided, then the invoker is ReadCrosshair
      * @return (narration, currentQuery):
-     *      "narration" is the actual one to be narrated through Narrator,
-     *      "currentQuery" is kind of shortened "narration" that is used for checking if target is changed compared to previous.
+     * "narration" is the actual one to be narrated through Narrator,
+     * "currentQuery" is kind of shortened "narration" that is used for checking if target is changed compared to previous.
```
sgmedev: This formatting change looks unrelated, same below.

